PR #13, opened 2024-09-17
Brodey/mem

brodeynewman:


server.cu
```diff
@@ -137,6 +137,37 @@ int request_handler(int connfd)
...
+        nvmlMemory_v2_t memoryInfo;
+
+        // Read the device handle from the client
+        std::cout << "Reading device handle from client" << std::endl;
+        if (read(connfd, &device, sizeof(nvmlDevice_t)) < 0) {
+            std::cerr << "Failed to read device handle" << std::endl;
+            return -1;
+        }
+
+        memoryInfo.version = nvmlMemory_v2;
```
kevmo314, line 152: Do a `if (read(connfd, &memoryInfo, sizeof(nvmlMemory_v2_t)) {` and read this from the client. We don't want to manipulate the client's request because we technically aren't API-compatible if we do.

kevmo314 (approved):

PR #15, opened 2024-09-18
chore: umap for funcs

brodeynewman:


server.cu
```diff
@@ -18,2040 +18,1809 @@
...
 
-        return result;
-    }
+    return result;
+}
 
-    case RPC_nvmlDeviceValidateInforom:
-    {
-        nvmlDevice_t device;
+static std::unordered_map<unsigned int, RequestHandler> opHandlers = {
```
kevmo314, line 1659: Try making this an array instead? the `#define`'s are all sequential so we should be able to merge them into one bigass array which would be faster than a map.

PR #20, opened 2024-09-23
chore: torch stuff after merge

brodeynewman:


client.cpp
```diff
@@ -154,13 +179,16 @@ int rpc_wait_for_response(const unsigned int request_id)
     }
 }
 
-int rpc_end_request(void *result, const unsigned int request_id)
+template <typename T>
```
kevmo314, line 182: `rpc_end_request` needs to be templated instead, `rpc_get_return` was removed.

kevmo314 (approved):

PR #43, opened 2024-10-27
Add basic connection logging to client / server

K0IN:
This PR adds a build flag -DVERBOSE=1 to log connection infos.

Infos logged:

Server:
* When a client connects

Client:
* When getaddrinfo fails
* When connects fails
* RPC__START fails


kevmo314:
Thank you! This looks like a great direction to take our logging.

PR #45, opened 2024-11-01
Brodey | RPC ordering

brodeynewman:
This PR introduces a `@disabled` doc to the annotations.

We can use this to enable the RPC call (which allows us to maintain static RPC ordering) while stubbing a 'disabled' server function.

This also works for manually implemented functions; any manual functions should exist in the annotations file with a @disabled flag.

Validated that tests are still passing with these changes.


codegen/annotations.h
```diff
@@ -1,17 +1,23 @@
 #include <nvml.h>
 #include <cuda.h>
 #include <cuda_runtime.h>
-
+/**
+ * @disabled
+ */
 nvmlReturn_t nvmlInit_v2();
```
kevmo314, line 7: This and shutdown should still need to be generated?

kevmo314 (approved):

PR #48, opened 2024-11-08
Brodey | decompress ptx

brodeynewman:


codegen/manual_client.cpp
```diff
@@ -69,7 +179,7 @@ cudaError_t cudaMemcpy(void *dst, const void *src, size_t count, enum cudaMemcpy
             return cudaErrorDevicesUnavailable;
         }
 
-        if (rpc_read(0, dst, count) < 0)
+        if (rpc_read(0, dst, sizeof(count)) < 0)
         {
```
kevmo314, line 183: This should be just `count`

PR #60, opened 2024-12-05
Working cublas batched example

brodeynewman:


codegen/codegen.py
```diff
@@ -403,12 +395,20 @@ class OpaqueTypeOperation:
     def client_rpc_write(self, f):
         if not self.send:
             return
-        f.write(
-            "        rpc_write(0, &{param_name}, sizeof({param_type})) < 0 ||\n".format(
-                param_name=self.parameter.name,
-                param_type=self.type_.format(),
+        elif "const double*" in self.type_.format():
```
kevmo314, line 398: Should this get mapped to a nullable type instead? An opaque type that is dereferenced is a bit odd.

PR #62, opened 2024-12-08
Brodey | unified memory

brodeynewman:


client.cpp
```diff
@@ -42,8 +52,76 @@ int nconns = 0;
...
+static void* faulting_address = nullptr;
+
+static void segfault(int sig, siginfo_t* info, void* unused) {
+    faulting_address = info->si_addr; 
+
+    std::cout << "Caught segfault at address: " << faulting_address << std::endl;
+
+    for (const auto & [ ptr, sz ] : conns[0].unified_devices)
+    {
+        if (faulting_address >= (ptr) && faulting_address <= (ptr + sz))
```
kevmo314, line 66:
```suggestion
        if (ptr <= faulting_address && faulting_address < (ptr + sz))
```

client.cpp
```diff
@@ -42,8 +52,76 @@ int nconns = 0;
 
 const char *DEFAULT_PORT = "14833";
 
+static int init = 0;
+static jmp_buf catch_segfault;
+static void* faulting_address = nullptr;
+
+static void segfault(int sig, siginfo_t* info, void* unused) {
```
kevmo314, line 59: Can you use `void* unused` to pass the parent segfault handler?

client.cpp
```diff
@@ -220,6 +298,54 @@ int rpc_read(const int index, void *data, size_t size)
...
+
+            // ptr is the same on both host/device
+            cudaError_t res = cudaMemcpy(ptr, ptr, size, cudaMemcpyDeviceToHost);
+
+            if (res != cudaSuccess) {
+                std::cerr << "cudaMemcpy failed :" << cudaGetErrorString(res) << std::endl;
+            } else {
+                std::cout << "Successfully copied " << size << " bytes" << std::endl;
+            }
+        }
```
kevmo314, line 331:
```suggestion
            size_t size = reinterpret_cast<size_t>(sz);

            // ptr is the same on both host/device
            cudaError_t res = cudaMemcpy(ptr, ptr, size, kind);
            if (res != cudaSuccess) {
                std::cerr << "cudaMemcpy failed :" << cudaGetErrorString(res) << std::endl;
            } else {
                std::cout << "Successfully copied " << size << " bytes" << std::endl;
            }
```

codegen/manual_client.cpp
```diff
@@ -792,3 +799,44 @@ extern "C"
...
+            rpc_wait_for_response(0) < 0 ||
+            rpc_end_response(0, &return_value) < 0)
+            return cudaErrorDevicesUnavailable;
+    } else {
+        if (rpc_start_request(0, RPC_cudaFree) < 0 ||
+            rpc_write(0, &devPtr, sizeof(void*)) < 0 ||
+            rpc_wait_for_response(0) < 0 ||
+            rpc_end_response(0, &return_value) < 0)
+            return cudaErrorDevicesUnavailable;
+    }
```
kevmo314, line 820: Can this be one block instead?
```suggestion
    maybe_free_unified_mem(0, devPtr);

        if (rpc_start_request(0, RPC_cudaFree) < 0 ||
            rpc_write(0, &devPtr, sizeof(void*)) < 0 ||
            rpc_wait_for_response(0) < 0 ||
            rpc_end_response(0, &return_value) < 0)
            return cudaErrorDevicesUnavailable;
```

kevmo314 (approved):

PR #63, opened 2024-12-09
fix: wait for stream to synchronize before freeing

kevmo314:
Fixes `cudaMemcpyAsync`.


kevmo314:
Tests pass locally

```
Running test(s)...
   ✓ CUDA is available.
   ✓ Tensor moved to CUDA successfully.
   ✓ Tensor successfully moved to CUDA and back to CPU.
   ✓ CUDA vector_add example works.
   ✓ cuDNN correctly applies sigmoid activation on a tensor.
   ✓ Batched cublas works via test/cublas_batched.cu.
(venv) kevin@inferable-node-006:~/scuda$ 
```

PR #65, opened 2024-12-23
Brodey | dynamic unified ptrs

brodeynewman:


codegen/codegen.py
```diff
@@ -209,7 +212,24 @@ def client_rpc_write(self, f):
...
-            
+
+    def client_unified_copy(self, f, direction):
+        f.write("    maybe_copy_unified_arg(0, (void*){name}, {direction});\n".format(name=self.parameter.name, direction=direction))
+
+        if isinstance(self.length, int):
+            f.write("    for (int i = 0; i < {name}; i++)\n".format(name=self.length))
+            f.write("       maybe_copy_unified_arg(0, (void*)&{name}[i], {direction});\n".format(name=self.parameter.name, direction=direction))
+        else:
+            if hasattr(self.length.type, "ptr_to"):
```
kevmo314, line 223:
```suggestion
        elif hasattr(self.length.type, "ptr_to"):
```

Flatten these

client.cpp
```diff
@@ -96,6 +96,28 @@ static void segfault(int sig, siginfo_t* info, void* unused) {
     raise(SIGSEGV);
 }
 
+void maybe_copy_unified_arg(const int index, void* arg, enum cudaMemcpyKind kind)
```
kevmo314, line 99: Make this return -1 if any cudaMemcpy fails? We probably want to return early in the generated rpc's if the memcpy's explicitly fail.

kevmo314 (approved):

PR #104, opened 2025-04-14
clear the warnings:

gaojs:
codegen/manual_client.cpp:125:37: warning: format '%x' expects argument of type 'unsigned int', but argument 2 has type 'size_t' {aka 'long unsigned int'}
codegen/manual_client.cpp:126:33: warning: format '%x' expects argument of type 'unsigned int', but argument 2 has type 'long long unsigned int'


kevmo314 (approved):

kevmo314:
Thank you!

PR #106, opened 2025-04-14
reorganize the included header files in the following order: C langua…

gaojs:
reorganize the included header files in the following order: C language, C++, system libraries, custom.
codegen.py will generate gen_api.h/gen_client.cpp/gen_server.cpp according cuda.h/cudnn.h/nvml.h...


kevmo314:
This is no longer necessary as we are stubbing at the driver level now.

PR #107, opened 2025-04-14
clear the warning: variable "size" is used before its value is set

gaojs:
codegen/gen_server.cpp(5531): warning #549-D: variable "size" is used before its value is set
      vgpuTypeClass = (char*)malloc(size * sizeof(char));

ref: https://docs.nvidia.com/deploy/nvml-api/group__nvmlVgpu.html nvmlVgpuTypeGetClass(nvmlVgpuTypeId_t vgpuTypeId, char* vgpuTypeClass, unsigned int* size), size is in&out param


kevmo314:
This has been addressed at some point.

PR #108, opened 2025-04-14
clear the warning: variable "dependencies" is used before its value is set

gaojs:
…arning:

codegen/gen_server.cpp(9786): warning #549-D: variable "dependencies" is used before its value is set
          rpc_read(conn, dependencies, numDependencies * sizeof(cudaGraphNode_t)) < 0 ||


kevmo314:
This has been addressed at some point.

PR #115, opened 2026-05-15
Fix segfault on LD_PRELOAD (use fprintf)

### LINKED ISSUE #101: Segfault when preloading scuda client libscuda_12.2.so even though the server is running
There is always an Segmentation Fault when preloading the client at ../libscuda_12.2.so whether local or remote, also the SCUDA_SERVER ip is set correctly as you can see.

root@cuda-gpu-worker6-scuda:/home/ubuntu/scuda/deploy# ./start.sh torch
../libscuda_12.2.so
Connecting to SCUDA server at: localhost:14833
Using scuda binary at path: ../libscuda_12.2.so
Running torch example...
./start.sh: line 23: 348232 Segmentation fault      (core dumped) LD_PRELOAD="$libscuda_path" python3 -c "import torch; print('CUDA Available:', torch.cuda.is_available())"

weshouman:
`std::cout`/`std::cerr` can be called before C++ streams are initialized when the library is loaded via `LD_PRELOAD`, because `dlsym()` is invoked during C++ runtime init itself. C stdio does not have this static initialization order dependency.

Fixes #101

### DIFF
--- a/client.cpp
+++ b/client.cpp
@@ -4,7 +4,6 @@
 #include <cuda_runtime.h>
 #include <dlfcn.h>
 #include <functional>
-#include <iostream>
 #include <map>
 #include <netdb.h>
 #include <netinet/tcp.h>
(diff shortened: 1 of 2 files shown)


kevmo314:
See https://github.com/lupinemachines/lupine/issues/101#issuecomment-4598205621 this should no longer be an issue.

PR #161, opened 2026-06-11
Codegen NVML client/server wrappers

kevmo314:
## Summary
- generate NVML client/server wrappers for the supported NVML surface
- return zero-valued success locally for NVML ECC counter queries used by nvidia-smi on T4 workers
- keep server-side ECC handlers compatible for older clients by draining request args and returning zero counters

## Verification
- cmake -S . -B build-codex -DCMAKE_BUILD_TYPE=RelWithDebInfo
- cmake --build build-codex -j24
- mounted the built libnvidia-ml.so.1 into ghcr.io/lupinemachines/lupine-client:cuda-13.1.0-ubuntu24.04 and verified full nvidia-smi completes through LUPINE_SERVER=demo.lupinemachines.com:14833 within the 15s demo window


kevmo314:
Closing this after rebasing onto current main and removing the zero-stub/manual ECC override. Current main already codegens these NVML ECC calls and returns the actual NVML return code plus output value, which is the correct behavior.

PR #206, opened 2026-06-25
Coalesce HTTP/2 frame writes into a single vectored send

kevmo314:


kevmo314:
Closing: benchmarked on a GPU node (driver-API microgpt-pattern A/B, loopback) and the syscall/packet reduction nets ~0% end-to-end — the saved send() overhead is dwarfed by RPC dispatch + round-trip cost on a low-latency link. Not worth the added lines on perf grounds. May revisit for real-WAN packet-rate reduction if measured across two machines.

kevmo314:
Reopening — the original 0% was measured on loopback, which has no RTT or delayed-ACK cost so request fragmentation is free there. Re-tested with the **correct topology** (CPU-only client → remote GPU server over a real ~1 Gbps LAN):

**Root cause:** `h2_write_all` sent each iovec as its own `send()`, so every RPC request went out as **4 separate TCP segments** (`[9B HTTP/2 header][4][4][4]`). Confirmed via /proc/net/snmp: **4.0 segments/RPC** each direction. Coalescing makes it **1.0 segment/RPC**.

**Effect on per-RPC latency** (cuMemcpyDtoH round-trip, 6 interleaved trials of 3000 RPCs each):

| | median | mean |
|---|---|---|
| main | ~424 µs | ~1410 µs |
| this PR | ~359 µs | ~450 µs |

Median **−15%**, mean **~3×** lower — the fragmented requests were triggering delayed-ACK/Nagle interactions that produced a heavy latency tail (p99 ~2.1 ms → often <1 ms). This directly speeds up every synchronizing RPC, which is what dominates latency-bound workloads like the microgpt training step.

PR #207, opened 2026-06-26
Route CUDA wrappers by object owner

kevmo314:
## Summary
- route cuLaunchHostFunc and cuStreamAddCallback through the stream/current-context owner instead of always connection 0
- add lupine_call_local_cuda_if_routed and use it from manual callback wrappers
- update codegen so generated CUDA client wrappers use the same helper instead of open-coding local dispatch
- add an integration regression that schedules both callback APIs on a non-default device stream

## Verification
- cmake --build build --parallel --target lupine_driver lupine_driver_server lupine_nvml
- SERVER_PORT=15127 TEST_TIMEOUT=180 ./test/run_custom_tests.sh
- two-server Lupine: LUPINE_SERVER=inferable-node-006:15128,inferable-node-008:15129 /tmp/lupine-routing-test/test_stream_callback_route
- local-device Lupine route on inferable-node-008 with shim loaded and no LUPINE_SERVER

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -55,6 +55,7 @@ set(CLIENT_SOURCES
     ${CMAKE_CURRENT_SOURCE_DIR}/rpc.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/compress.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/client.cpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/client_routing.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/codegen/gen_client.cpp
 )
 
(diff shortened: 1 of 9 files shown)


kevmo314:
Follow-up from review: I added a typed helper, `lupine_call_local_cuda_if_routed`, so manual wrappers do not need to open-code the local-route branch. This cannot live purely in the RPC transport because local dispatch needs the CUDA ABI symbol and typed arguments, which the transport layer no longer has. Generated wrappers already get local-route handling from `codegen/codegen.py`; this PR fixes the manual-wrapper gap without regenerating the whole client.

kevmo314:
Updated per review: codegen now emits `lupine_call_local_cuda_if_routed` into `gen_client.cpp` and generated wrappers call it instead of open-coding the local route branch. The manual callback wrappers use the same helper shape. Rebuilt and reran custom tests plus the two-server route regression.

PR #208, opened 2026-06-26
Reuse a pinned staging buffer for host-to-device copies

kevmo314:


kevmo314:
Found by profiling: the server called `cuMemAllocHost`+`cuMemFreeHost` on **every** HtoD, page-locking and unlocking host memory per call (perf showed the server child spending ~all its time in `__handle_mm_fault` / `shmem_alloc_and_add_folio` / `__get_user_pages` / `zap_pte_range`). For small/frequent copies that pinning churn, not the transfer, is the cost.

Fix: reuse a process-global pinned staging buffer (safe — the server forks one process per connection and runs its handlers serially). Buffers ≤8 MB are retained and reused; larger requests fall back to a per-call allocation whose cost is amortized over the big transfer.

Measured on the GPU node (CPU-only client → remote server):

| small `cuMemcpyHtoD` (64 B) | main | this PR |
|---|---|---|
| server-side (loopback, network removed) | 1156 µs/op | **52.7 µs/op (22×)** |
| end-to-end over ~1 Gbps LAN | ~2900 µs/op | **~1490 µs/op (~2×)** |

Correctness: HtoD→DtoH round-trip `memcmp` passes across 64 B → 100 MB, including reuse, buffer growth, reusing a large buffer for a small request, and the >64 MB multi-chunk + >8 MB per-call paths.

PR #209, opened 2026-06-26
Send async kernel launches and copies fire-and-forget

kevmo314:


kevmo314:
Profiling real eager microgpt training (torch, CPU-only client → remote GPU server) showed `cuLaunchKernel` was the dominant per-step cost: a synchronous RPC round-trip per launch, ~82/step. Real CUDA launches are asynchronous, so the ack is almost always unnecessary.

This makes async launches (`cuLaunchKernel`) and async device-to-device copies (`cuMemcpyDtoDAsync`) **fire-and-forget** — the client sends without waiting and returns `CUDA_SUCCESS`; the server executes without responding. Requests stay FIFO-ordered on the connection, so a later synchronizing RPC still observes them. Fire-and-forget requests are also **coalesced** into a per-connection buffer and flushed as one HTTP/2 DATA frame by the next response-bearing request, amortizing per-request framing + `sendmsg` + per-packet netfilter.

**Measured per training step** (loss identical to native at every config):

| model | params | native | baseline Lupine | this PR (+ #206/#208) |
|---|---|---|---|---|
| toy microgpt | 0.03M | 1.2 ms | ~67 ms | ~4 ms |
| L6 e384 | 11M | 53.7 ms | 395 ms (7.4×) | 58 ms (1.08×) |
| L12 e768 (GPT-2 small) | 86M | 337 ms | 854 ms (2.5×) | 339 ms (**1.005×**) |

The overhead is roughly fixed per step, so at realistic model scale it amortizes to ~native; the baseline penalty (which round-trips every launch) instead grows with layer count.

⚠️ **Caveat / follow-up:** this is a prototype — a launch/copy error is currently dropped rather than surfaced. CUDA's model is to report async launch errors at the next synchronization; the production version should have the server capture launch errors and return them from the next `cuCtxSynchronize`/`cuStreamSynchronize`. Also the `cuMemcpyDtoDAsync` edit is in the generated files (`gen_client.cpp`/`gen_server.cpp`) and should move into the codegen annotations so it survives regeneration.

kevmo314:
Correction on the error-handling caveat — it's smaller than I first wrote. **Sticky errors propagate automatically** and need no extra work: the server holds the CUDA context, so a sticky error (illegal memory access, device-side assert, etc.) poisons it and is returned by the next real CUDA call the server makes — i.e. the client's next synchronizing RPC. Verified through this PR's build: an out-of-bounds gather launched fire-and-forget surfaces correctly as `CUDA error: device-side assert triggered` at the next `torch.cuda.synchronize()`.

The only errors lost are **non-sticky synchronous launch-config errors** (invalid grid/block dims, `LAUNCH_OUT_OF_RESOURCES`) — caught at launch-issue time, not sticky. These don't occur in a working loop (valid configs) and are dev-time bugs, so dropping their synchronous report is acceptable. No deferred-error machinery required.

kevmo314:
Fixed the codegen CI failure. The `cuMemcpyDtoDAsync` fire-and-forget change had been hand-edited into the generated files (`gen_client.cpp`/`gen_server.cpp`), which the codegen regen-and-diff check rejects. Reverted those to the codegen output, so this PR now touches only non-generated code (`client.cpp`, `manual_server.cpp`, `rpc.cpp`, `rpc.h`).

The dropped D2D copy was marginal (~0.5 ms on the toy model, negligible at real scale) — the win is the `cuLaunchKernel` fire-and-forget + batching, which is intact. Extending fire-and-forget to async copies/memsets the right way (a codegen `@async` annotation so the generated client/server emit the no-wait/no-response path) is a clean follow-up.

kevmo314:
Added the `@async` codegen annotation (the proper fix for the earlier hand-edit). A function tagged `@async` in `annotations.h` now generates a fire-and-forget client (queues into the coalescing buffer via `rpc_write_start_request_async` + `rpc_write_end_batched`, returns `CUDA_SUCCESS` without waiting) and a server handler that executes the call and sends no response. Applied it to `cuMemcpyDtoDAsync_v2`, so the D2D copy is back — generated, not hand-edited.

Verified the codegen regenerates cleanly in a CI-equivalent environment (CUDA 13.1 + clang-format 18.1.3): a no-op regen reproduces the committed files byte-for-byte, and with `@async` the **only** generated change is the `cuMemcpyDtoDAsync` client/server (22 lines). microgpt loss still 3.6093. Other async copies/memsets can now opt in just by adding `@async` to their annotation.

kevmo314:
Per review: dropped the request batching/coalescing. Fire-and-forget ops now just send the request and return without waiting for the ack (server sends no response), instead of deferring into a per-connection buffer flushed at the next sync.

Two reasons: (1) it's simpler — `rpc.cpp`/`rpc.h` are now untouched, and the change is just the fire-and-forget send + the `@async` codegen; (2) it's better behavior — deferring delayed GPU execution (launches sat in the client buffer until the next sync, so the GPU couldn't start until the flush), whereas sending immediately streams each launch to the server so the GPU executes it right away, overlapping issue with compute like real async CUDA. The batching only saved syscalls on the pathological toy model and is negligible at real scale.

PR #225, opened 2026-07-07
fix(server): cap peer-supplied module image size

kevmo314:
## Problem
`handle_manual_cuModuleLoad` / `cuModuleLoadData` / `cuLibraryLoadData` read `image_size` straight off the wire and pass it to a `std::vector` allocation. A peer-supplied `image_size == SIZE_MAX` made `vector(image_size + 1)` wrap to a zero-length buffer, after which the server tried to read `SIZE_MAX` payload bytes and **hung the connection child forever**. The plain `vector(image_size)` variants threw `std::bad_alloc` (uncaught → `std::terminate`). Either way a single malformed peer wedged the connection.

## Fix
Cap `image_size` at 2 GiB (well above any real fat binary) and reject the request up front.

## Reproduction
`test/test_module_image_overflow.cpp` is a crafted client that links the lupine RPC transport directly and sends `cuModuleLoad` with `image_size = SIZE_MAX`, asserting the server closes the connection instead of hanging (5s `SIGALRM` watchdog).

Verified against a live server:
- **before fix:** `RESULT: HANG (no close within timeout -> vulnerable)` (exit 1)
- **after fix:** `RESULT: PASS (server rejected malformed request)` (exit 0)

Wired into `local.sh` (`test_module_overflow`) and built via CMake (`module_image_overflow_test`).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -139,6 +139,18 @@ else()
     target_link_libraries(h2_test PRIVATE stdc++ ${NGHTTP2_LIBRARY} lupine_lz4 pthread)
     enable_testing()
     add_test(NAME h2_test COMMAND h2_test)
+
+    # Robustness regression tests: crafted clients that speak the wire protocol
+    # directly. They need a running lupine_driver_server (LUPINE_SERVER) but no
+    # GPU, so they are built here but run by local.sh against the live server.
+    add_executable(module_image_overflow_test
+        ${CMAKE_CURRENT_SOURCE_DIR}/rpc.cpp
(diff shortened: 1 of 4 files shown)


kevmo314:
Closing — on reflection this isn't the server's job to police. The server forks a separate child process per connection, so a client that sends an oversized image_size only hangs/crashes its own child (blocked in recv, no CUDA context initialized yet, ~process-baseline memory). Other connections are unaffected and it's essentially zero server resource usage. The aggregate worst case is no different from any connection-flooding DoS the fork model already accepts. Thanks for the pushback.

Issue #291, opened 2026-07-10
Performance: cache decoded kernel parameter layouts


kevmo314:
## Opportunity

Kernel launch handling repeatedly derives parameter sizes/offsets/layout from module/function metadata. Stable kernels are launched many times, so repeating this work adds allocation, parsing, and driver-query overhead to the hot path.

## Proposed direction

Cache the validated parameter layout by a stable function/module identity and invalidate entries when modules or functions are unloaded/replaced. Bound the cache or tie it directly to module lifetime.

## Measurement and acceptance

- Add a repeated-launch benchmark with small kernels.
- Show cache hit/miss and invalidation behavior in tests.
- Preserve correctness for module unload/reload and identical symbol names in different modules.
- Report CPU time and launch latency before/after.

kevmo314:
I prototyped this on `45536ae` with a bounded, generation-aware cache keyed by `CUfunction`, global invalidation on successful module/library/context lifetime-ending calls, and a generation token that prevents an unload racing an in-flight miss from repopulating stale data.

Correctness coverage included hit/miss/invalidation/capacity unit tests and a GPU integration test with two same-named kernels with different layouts plus 32 alternating unload/reload cycles. CUDA reused the opaque function handle 26 times. This also exposed a separate pre-existing client-cache invalidation bug: the baseline test failed (`expected 8, got 0`) while explicit client invalidation passed.

Performance result: **I do not think the server cache is worth merging in its current form.**

- Diagnostic run, 50,000 measured launches + 1,000 warmups: cache off made 51,001 layout derivations, spending 35.3–43.5 ms total inside `cuFuncGetParamInfo`; cache on made one derivation (0.018–0.020 ms) and recorded 51,000 hits.
- Decisive latency run used separate otherwise-identical servers with diagnostic timing/counters disabled, loopback transport on the GPU host, 8 interleaved runs per mode, and 50,000 measured launches/run.
- Median launch wall time: 32.158 us off vs 32.085 us on (0.23% difference). Paired deltas straddled zero.
- Median client CPU time: 24.108 us off vs 24.137 us on (no improvement).
- WAN measurements were substantially noisier and did not resolve a signal.

[...]

PR #319, opened 2026-07-10
Invalidate kernel layout caches on CUDA object teardown

kevmo314:
## Summary

- invalidate cached kernel layouts after successful module, library, and context lifetime-ending calls on both local and remote routes
- use a lock-free 128-slot direct-mapped `thread_local` layout cache with a global atomic invalidation epoch
- keep private cache storage/hash/epoch logic in client-only `cache.cpp`, with `cache.h` exposing only the layout value and narrow lookup/insert/invalidate functions; no cache wrapper class or new dependency
- add a GPU regression covering same-named kernels with different layouts, cross-thread invalidation, and unload/reload handle reuse

## Cache semantics

Each participating thread owns 128 fixed slots (about 132 KiB/thread). A slot is usable only when its stored epoch matches the global atomic epoch. Teardown increments that epoch; the next lookup on every thread rejects old slots. If teardown races an in-flight RPC fetch, the fetch can only write the old epoch and therefore cannot become a later hit. Direct-map collisions cause a safe refetch, never a stale result.

This intentionally provides the eventual consistency CUDA requires for concurrent destroy/use races without a shared-map mutex or unsynchronized shared container.

## Regression proof

Before this change, the cross-thread integration test fails against `origin/main` after CUDA reuses a function handle with a different parameter layout:

```
kernel output mismatch: expected 109, got 0
```

With this change, the persistent worker-thread cache is invalidated by main-thread unloads and the test passes; the final run observed 19 cross-layout handle reuses.

## Validation

- `cmake --build build -j2`
- `SERVER_HOST=inferable-node-006 SERVER_PORT=15303 test/run_custom_tests.sh` — 9/9 passed
- 50k-launch loopback benchmark: mutex cache 31.698 us/launch, thread-local cache 31.782 us/launch (no measurable regression)

## Performance follow-up

Related to #291. I also prototyped the proposed server-side layout cache, but a statistics-free 8x50k-launch A/B showed no credible performance benefit (32.158 us/launch cache-off vs 32.085 us/launch cache-on, paired samples straddling zero), so that speculative portion is intentionally excluded. Full measurements are documented on #291.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -56,6 +56,7 @@ set(CLIENT_SOURCES
     ${CMAKE_CURRENT_SOURCE_DIR}/h2.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/rpc.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/compress.cpp
+    ${CMAKE_CURRENT_SOURCE_DIR}/cache.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/client.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/client_routing.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/codegen/gen_client.cpp
@@ -79,6 +80,7 @@ set(NVML_CLIENT_SOURCES
 )
 
 set(CLIENT_HEADERS
+    ${CMAKE_CURRENT_SOURCE_DIR}/cache.h
     ${CMAKE_CURRENT_SOURCE_DIR}/client_routing.h
     ${CMAKE_CURRENT_SOURCE_DIR}/codegen/gen_client.h
 )
@@ -144,6 +146,7 @@ else()
         ${CMAKE_CURRENT_SOURCE_DIR}/h2_test.cpp
     )
     target_link_libraries(h2_test PRIVATE stdc++ ${NGHTTP2_LIBRARY} lupine_lz4 pthread)
+
     enable_testing()
     add_test(NAME h2_test COMMAND h2_test)
     add_test(
--- a/cache.cpp
+++ b/cache.cpp
@@ -1,0 +1,85 @@
+#include "cache.h"
+
+#include <array>
+#include <atomic>
+#include <functional>
+
+namespace {
+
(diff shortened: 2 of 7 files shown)


kevmo314:
Follow-up audit on reusing an existing thread-safe map:

There is no concurrent/thread-safe map abstraction on `main`. PR #314 (still open) introduces `lupine_handle_registry`, but its API is snapshot `lookup`, unconditional `store`, `remove`, and `get_or_create`; it does not provide whole-cache invalidation epochs, bounded capacity, or conditional insertion against a captured epoch.

That distinction is required here because the layout fetch is an RPC and must run outside the map lock. If an unload clears the cache while that RPC is in flight, a normal thread-safe `store()` can repopulate the now-stale layout afterward. `generation_cache` captures the epoch at miss time and rejects that post-invalidation insertion. A TBB-style concurrent map would have the same semantic gap and would add a new dependency.

So I have not churned the PR: there is no existing utility that preserves the required handle-reuse/invalidation behavior, and folding these semantics into the pending server-handle registry would couple unrelated PRs while effectively recreating the same generation-aware cache.

kevmo314:
Updated per the lock-free/fewer-wrapper follow-up in `39fd20f`:

- removed `generation_cache.h`, its standalone test, and the CMake target
- replaced the shared mutex/map with a global atomic epoch and 128 fixed direct-mapped `thread_local` slots in `client.cpp`
- added cross-thread regression coverage: a persistent worker owns the cached layout while the main thread unloads/reloads modules; untouched main fails (`expected 109, got 0`), updated code passes with 19 cross-layout handle reuses
- bounded memory is ~132 KiB per participating thread; collisions only trigger RPC refetch
- 50k single-thread launch benchmark remains tied (31.698 us mutex vs 31.782 us thread-local)
- CTest 2/2 and GPU custom suite 9/9 pass

kevmo314:
Moved the lock-free cache implementation to the requested future cache home in `382acae`: private epoch/thread-local slots/key/hash are in client-only `cache.cpp`; `cache.h` contains only the layout value and narrow API. `client.cpp` now only coordinates fetch/use and calls invalidation. Rebuild, CTest 2/2, and the cross-thread GPU regression (19 cross-layout reuses) pass after the move.

PR #321, opened 2026-07-10
Receive HTTP/2 payloads directly into RPC destinations

### LINKED ISSUE #293: Performance: read HTTP/2 payloads directly into RPC destination buffers
## Opportunity

Inbound HTTP/2 data is first copied into transport-owned queued storage and then copied again by `rpc_read()` into the caller's destination. Large host/device transfers therefore pay an avoidable memory allocation and full extra memory copy.

## Proposed direction

Allow the active RPC reader to provide destination spans to the HTTP/2 data callback, with bounded staging only for headers, unmatched responses, or fragmented framing. Carefully preserve multiplexing and lifetime rules.

## Measurement and acceptance

- Add copy/allocation counters and a large-transfer benchmark.
- Demonstrate one fewer full-payload copy on the direct path.
- Cover fragmented frames, partial reads, concurrent lanes, errors, and compressed payloads.
- Report throughput and peak resident memory before/after.

kevmo314:
## Summary

- let the active synchronous RPC reader lend its remaining destination span to the nghttp2 DATA callback
- copy matching DATA directly into that span and stage only overflow/read-ahead that arrives before the next destination is known
- install the borrowed destination only while `nghttp2_session_mem_recv()` runs under the session mutex, and clear it before every return
- reject unexpected HTTP/2 stream IDs rather than sharing one destination across streams
- retain transport counters for direct bytes, staged bytes/copies, staging buffers, and peak staged storage

Direct receive is always enabled; there is no runtime fallback branch.

Closes #293.

## Lifetime and multiplexing

Lupine currently uses one long-lived HTTP/2 stream per connection. RPC request IDs and lanes multiplex messages within that byte stream, but `read_id` reserves exactly one inbound body until `rpc_read_end()`. The destination pointer is therefore borrowed only for the synchronous `rpc_http2_read()` call and only while nghttp2 is consuming the just-read 64 KiB socket buffer. It is cleared before releasing the session mutex on success or error.

A small header read can still cause the rest of that socket buffer to arrive before the payload destination is known. That tail remains in the existing queue; counters and tests show read-ahead stays bounded to 64 KiB.

## Correctness coverage

`h2_test` covers:

- fragmented DATA frames spanning one destination
- partial reads that direct-copy 7 bytes and stage/copy exactly the 4,089-byte overflow
- peer close after partial delivery with destination guard-byte checks
- two concurrent RPC waiters receiving 1 MiB responses sent in reverse request-ID order
- mixed compressed and raw LZ4 framing through the direct path
- unexpected HTTP/2 streams are rejected rather than sharing a global destination

## Benchmark results

Development A/B using Release builds on two pinned CPUs, 64 MiB payload x 24 measured iterations (1.5 GiB/sample), eight interleaved samples/mode with a short recovery interval:

| mode | median wall | throughput | median process CPU | payload counters |
| --- | ---: | ---: | ---: | --- |
| staged baseline | 0.444 s | 3.38 GiB/s | 0.791 s | 1,677,721,601 staged + copied bytes; ~28k staging buffers; 64 KiB peak |
| direct | 0.387 s | 3.88 GiB/s | 0.711 s | 1,677,721,601 direct bytes; 0 staged bytes/buffers |

That is +14.7% throughput and -10.2% process CPU. Peak RSS was effectively unchanged (~135 MiB) because source/destination buffers dominate and the old staging queue was already bounded to one 64 KiB socket read.

GPU loopback benchmark on node006, 64 MiB incompressible payload x 8, eight interleaved samples/mode:

- HtoD: 0.897 vs 0.899 GiB/s median (neutral); client CPU 0.1178 vs 0.1166 s
- DtoH: 1.804 vs 1.783 GiB/s median (within noise); client CPU 0.0781 vs 0.0737 s (-5.6%)
- peak client RSS unchanged at ~142 MiB

The realistic throughput is dominated by LZ4 raw-fallback work and GPU transfer; the isolated transport result shows the removed allocation/copy cost directly. Benchmark source files are intentionally not included in this PR.

## Validation

- `cmake --build build -j2`
- `SERVER_HOST=inferable-node-006 SERVER_PORT=15304 test/run_custom_tests.sh` — 9/9 passed

### DIFF
--- a/h2.cpp
+++ b/h2.cpp
@@ -37,6 +37,10 @@ struct h2_transport {
   int response_status = 0;
   nghttp2_session *session = nullptr;
   std::deque<h2_buffer> local_out;
+  unsigned char *read_destination = nullptr;
+  size_t read_remaining = 0;
+  rpc_http2_read_stats read_stats = {};
+  uint64_t staged_bytes = 0;
   // Reusable scratch holding the one LZ4-framed payload block currently in
   // flight (see h2_materialize_block). Writes are serialized per connection,
   // so a single buffer suffices and memory stays bounded by one block.
@@ -73,14 +77,31 @@ struct h2_write_source {
   }
 };
 
-void queue_bytes(std::deque<h2_buffer> &queue, const unsigned char *data,
-                 size_t len) {
+void receive_bytes(h2_transport *transport, const unsigned char *data,
+                   size_t len) {
+  if (len == 0) {
+    return;
+  }
+  size_t direct = std::min(len, transport->read_remaining);
+  if (direct != 0) {
+    memcpy(transport->read_destination, data, direct);
+    transport->read_destination += direct;
+    transport->read_remaining -= direct;
+    transport->read_stats.direct_bytes += direct;
+    data += direct;
+    len -= direct;
+  }
   if (len == 0) {
     return;
   }
   h2_buffer buffer;
   buffer.data.assign(data, data + len);
-  queue.push_back(std::move(buffer));
+  transport->local_out.push_back(std::move(buffer));
+  transport->read_stats.staged_bytes += len;
+  ++transport->read_stats.staged_buffers;
+  transport->staged_bytes += len;
+  transport->read_stats.peak_staged_bytes = std::max(
+      transport->read_stats.peak_staged_bytes, transport->staged_bytes);
 }
 
 // Maximum buffers per vectored send. Frames carry far fewer iovecs than this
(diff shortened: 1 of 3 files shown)


kevmo314:
Updated in `25a4446` per review: direct receive is now unconditional; removed `LUPINE_HTTP2_DIRECT_READ`, the fallback branch/test, both benchmark source files, and all benchmark CMake wiring. Observability counters and direct-path correctness tests remain. Rebuild and CTest 2/2 pass.

Issue #292, opened 2026-07-10
Performance: pipeline compression, network, and PCIe transfer stages


kevmo314:
## Opportunity

Large transfers are largely staged serially: prepare/compress a block, transmit/receive it, then perform the host-device copy. This leaves CPU, network, and PCIe engines idle while another stage runs.

## Proposed direction

Use bounded pinned-memory double buffers (or a small configurable ring) to overlap compression/decompression, HTTP/2 I/O, and asynchronous CUDA copies. Apply backpressure and avoid unbounded pinned allocations.

## Measurement and acceptance

- Prototype with configurable block/ring size and a safe fallback.
- Validate data integrity and stream-ordering semantics.
- Benchmark HtoD and DtoH over representative bandwidth/latency links and payload sizes.
- Report throughput, CPU use, pinned-memory footprint, and the crossover point where pipelining helps.

## Subissues

- [ ] #323 — async HtoD pinned-ring fast path (P0)
- [ ] #324 — pipeline synchronous HtoD receive and DMA (P1)
- [ ] #325 — double-buffer synchronous DtoH and transport (P1)
- [ ] #326 — eager async DtoH delivery (P2a)
- [ ] #328 — bounded async DtoH staging design (P2b)
- [ ] #327 — overlap LZ4 production and HTTP/2 writes (P3)

Benchmark infrastructure remains tracked in #302. The inbound HTTP/2 copy removal is implemented separately in #321 / #293, and the async HtoD cleanup bug is handled by #320.

kevmo314:
Investigation update (benchmarked on `origin/main`, with `LUPINE_DISABLE_LOCAL=1` on GPU client hosts so ordinal 0 is the remote device).

### What is already pipelined

The framing layer does not materialize a full compressed payload. It LZ4-compresses one 4 MiB block as nghttp2 pulls data, sends it, then reuses the scratch buffer. The receive side likewise decodes one block at a time.

The async HtoD handler also begins `cuMemcpyHtoDAsync` after each received 64 MiB chunk, so GPU DMA can overlap receipt of later chunks. Its problem was that it pinned the entire payload and attempted to free it from a CUDA host function. That concrete leak is fixed separately in #320.

### Baseline results

All numbers are effective GiB/s for 256 MiB payloads.

| path | link/workload | sync HtoD | sync DtoH | async HtoD + sync | async DtoH + sync |
|---|---|---:|---:|---:|---:|
[...]

PR #330, opened 2026-07-10
Pipeline synchronous HtoD transfers

### LINKED ISSUE #324: Performance: pipeline synchronous HtoD receive and device copy
Part of #292. This should reuse the bounded staging primitive from the async HtoD ring issue.

## Evidence

On node006, 256 MiB synchronous HtoD takes 167.1 ms (1.496 GiB/s), versus 85.4 ms (2.929 GiB/s) natively. handle_manual_cuMemcpyHtoD_v2 currently performs receive/decompress, blocking cuMemcpyHtoD_v2, then the next receive. Its 64 MiB chunk also exceeds the 8 MiB retained threshold, so each call pays a measured ~32.2 ms pin/unpin lifecycle.

## Proposed prototype

- Reuse two bounded pinned slots instead of allocating a 64 MiB buffer per call.
- Enqueue each received slot while receiving/decompressing the next.
- Wait for final completion before replying.
- Preserve synchronous driver ordering, including legacy-default-stream interactions; do not use an independent stream without explicit dependencies.
- Retain payload-draining behavior after a CUDA error.

## Correctness coverage

- work before and immediately after the copy on legacy default, blocking, and non-blocking streams
- multiple host threads sharing a context
- invalid destinations and failures after successful chunks
- zero-byte and boundary-sized payloads
- context switch/destruction while active

## Measurement and acceptance

- split wall time into allocation, receive/decode, DMA, and waits
- benchmark the #292 size/compressibility/link matrix
- target 2.7 GiB/s or better for 256 MiB loopback after warm-up, subject to ordering correctness
- bounded pinned memory and no slow-link regression

kevmo314:
## Summary

- pipeline synchronous HtoD receives through two retained 8 MiB pinned slots
- enqueue DMA on `CU_STREAM_LEGACY` while the next slot is received/decompressed
- wait for per-slot disable-timing events before reuse and before replying
- preserve the original serial handler as a no-consumption allocation/setup fallback
- quarantine the fast-path pool after uncertain transport or CUDA completion failures
- make the retained serial staging cache context-aware and contention-safe
- retire context-owned pinned staging before context destroy/detach and primary-context reset/release

Closes #324.

## Ordering and failure semantics

Using `CU_STREAM_LEGACY` preserves the ordering of a synchronous CUDA copy with prior work in blocking streams without synchronizing unrelated non-blocking streams. Final event waits preserve host-synchronous completion.

The handler continues draining unread framed payload after CUDA failures. If event setup fails before reading any payload, it uses the old serial path. If event recording fails after a copy was accepted, it synchronizes the legacy stream and finishes the unread suffix through the serial path without exposing the internal event error as the memcpy result. A copy-call error quarantines the pool because CUDA may be surfacing a deferred error and cannot prove that the just-read slot is unused.

The fast pool retains 16 MiB of portable pinned memory per server process. The existing serial fallback may retain another 8 MiB.

## Benchmark

Corrected physical-GPU control on `inferable-node-008`: RTX 2070 SUPER selected with `CUDA_DEVICE_ORDER=PCI_BUS_ID CUDA_VISIBLE_DEVICES=0`, plaintext loopback, 256 MiB incompressible/pageable synchronous HtoD, 5 warmups plus 15 measured copies, and three fresh-server repeats:

|---|---:|---:|
| exact branch base (`76cdcbe1`) | 116.207 ms | 2.151 GiB/s |
| this PR (`c09e8070`) | 63.064 ms | 3.964 GiB/s |
| native CUDA | 20.724 ms | 12.063 GiB/s |

The pipeline is 1.843x baseline throughput (+84.3%) and reaches 32.9% of native. Candidate repeats were 4.272, 3.929, and 3.964 GiB/s.

An earlier revision of this description used `CUDA_VISIBLE_DEVICES=0` without setting CUDA's device order. FASTEST_FIRST mapped that ordinal to the busy physical RTX 4090, so the old native-rate percentage was invalid; the table above supersedes it. The positive pipeline gain remains substantial under the corrected control.

## Validation

- Release build
- `CUDA_VISIBLE_DEVICES= SERVER_HOST=inferable-node-008 SERVER_PORT=15949 TEST_TIMEOUT=180 test/run_custom_tests.sh` — 11/11 passed
- focused coverage for size/framing boundaries, slot reuse, failed-copy payload draining, legacy blocking-stream ordering, non-blocking-stream exclusion, serial and pipelined context recreation, and context detach
- `git diff --check`

Part of #292.

### DIFF
--- a/codegen/codegen.py
+++ b/codegen/codegen.py
@@ -2076,6 +2076,7 @@ def main():
             '#include "gen_server.h"\n\n'
             '#include <cstdio>\n\n'
             '#include "rpc.h"\n\n'
+            '#include "manual_server.h"\n\n'
             '#include "nvml_server.h"\n\n'
         )
         for function, annotation, operations, metadata in functions_with_annotations:
@@ -2133,6 +2134,13 @@ def main():
                 for op in operations:
                     if op.parameter.name == param.name:
                         params.append(op.server_reference)
+
+            if function.name.format() in ("cuCtxDestroy_v2", "cuCtxDetach"):
+                f.write("    lupine_before_context_destroy(ctx);\n")
+            if function.name.format() == "cuDevicePrimaryCtxRelease_v2":
+                f.write("    lupine_before_primary_context_release(dev);\n")
+            if function.name.format() == "cuDevicePrimaryCtxReset_v2":
+                f.write("    lupine_before_primary_context_reset(dev);\n")
 
             if function.return_type.format() != "void":
                 f.write(
--- a/codegen/gen_server.cpp
+++ b/codegen/gen_server.cpp
@@ -18,6 +18,8 @@
 
 #include "rpc.h"
 
+#include "manual_server.h"
+
 #include "nvml_server.h"
 
 int handle_cuInit(conn_t *conn) {
@@ -490,6 +492,7 @@ int handle_cuDevicePrimaryCtxRelease_v2(conn_t *conn) {
   request_id = rpc_read_end(conn);
   if (request_id < 0)
     goto ERROR_0;
+  lupine_before_primary_context_release(dev);
   lupine_intercept_result = cuDevicePrimaryCtxRelease_v2(dev);
 
   if (rpc_write_start_response(conn, request_id) < 0 ||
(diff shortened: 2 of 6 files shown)


kevmo314:

PR #331, opened 2026-07-10
Pipeline synchronous DtoH transfers

### LINKED ISSUE #325: Performance: double-buffer synchronous DtoH copies and transport writes
Part of #292.

## Evidence

On node006, 256 MiB synchronous DtoH takes 137.5 ms (1.818 GiB/s), versus 81.9 ms (3.051 GiB/s) natively. handle_manual_cuMemcpyDtoH_v2 copies one 4 MiB chunk synchronously, then fully compresses/writes it before starting the next. The measured total is close to native DMA (~81.9 ms) plus remaining transport work (~55.6 ms), so overlap can plausibly approach native rate.

## Proposed prototype

- Use two bounded pinned 4–8 MiB slots.
- Enqueue the next DtoH copy while compression/network transmission consumes the completed slot.
- Track completion before sending or reusing a slot.
- Preserve existing per-chunk result/payload framing and partial-error behavior.
- Preserve legacy-default-stream and synchronous host-completion semantics.

## Correctness coverage

- integrity across block, frame, and slot boundaries
- ordering with legacy default, blocking, and non-blocking streams
- partial CUDA errors after emitted chunks
- disconnect/backpressure while a copy is in flight
- zero-byte and non-slot-aligned payloads

## Measurement and acceptance

- report DMA, wait, compression, socket-write, CPU, and pinned-memory measurements
- benchmark random/compressible 1 MiB–1 GiB on loopback and limited links
- target at least 80% of native loopback DtoH after warm-up
- compatible wire framing, bounded memory, and no slow-link regression

## Implementation notes from code audit

The smallest compatible state machine keeps the current 4 MiB multi-response framing:

1. Allocate two pinned slots and two disable-timing events.
2. Enqueue slot A on CU_STREAM_LEGACY and record event A.
3. Enqueue B before synchronizing/sending A.
4. While CPU/LZ4/socket consumes A, the GPU fills B; alternate until complete.
5. Wait every event, including the final slot, before send/reuse/return.

rpc_write_end consumes a slot synchronously, so it is reusable after that call returns. If setup fails before enqueue, fall back to the current serial path. If a later enqueue fails, finish/send the already-submitted slot and emit the failure result at the correct chunk position. If event recording fails after enqueue, synchronize the legacy stream before reuse and fall back to serial behavior rather than exposing the internal event error as the memcpy result.

kevmo314:
## Summary

- double-buffer synchronous DtoH through two pinned 4 MiB slots
- enqueue copies on `CU_STREAM_LEGACY` and record disable-timing events
- overlap the next DMA with compression and transport of the completed slot
- retain the existing 4 MiB multi-response framing and per-chunk CUDA results
- keep transfers below the measured 32 MiB crossover on the serial path
- resume the serial implementation at the first unsent byte after internal setup/event failures

Closes #325.

## Semantics and bounds

The legacy stream preserves synchronous-copy ordering with blocking streams. Every event is waited before its slot is sent or reused, including disconnect/error cleanup. Cleanup frees staging only after event or legacy-stream completion is confirmed; if both checks fail after an accepted DMA, the allocation is quarantined until context teardown and the CUDA failure is emitted at the next response chunk. CUDA memcpy failures are emitted at the same chunk position as the existing handler, and no later chunks are sent.

The pipeline allocates two 4 MiB pinned slots for transfers of at least 32 MiB and frees them before returning. Smaller transfers retain the allocation/event-free serial implementation.

## Benchmark

Corrected physical-GPU control on `inferable-node-008`: RTX 2070 SUPER selected with `CUDA_DEVICE_ORDER=PCI_BUS_ID CUDA_VISIBLE_DEVICES=0`, plaintext loopback, 256 MiB incompressible/pageable synchronous DtoH, 5 warmups plus 15 measured copies, and three fresh-server repeats:

|---|---:|---:|
| exact branch base (`23f748a6`) | 82.455 ms | 3.032 GiB/s |
| this PR (`5a97aa69`) | 68.080 ms | 3.672 GiB/s |
| native CUDA | 20.718 ms | 12.067 GiB/s |

The pipeline is 1.211x baseline throughput (+21.1%) and reaches 30.4% of native. Candidate repeats were 3.767, 3.231, and 3.672 GiB/s. A slow-link control remains effectively flat because both implementations already saturate that link.

An earlier revision of this description used `CUDA_VISIBLE_DEVICES=0` without setting CUDA's device order. FASTEST_FIRST mapped that ordinal to the busy physical RTX 4090, so the old native-rate percentage was invalid; the table above supersedes it. The DtoH overlap remains a positive improvement under the corrected control.

## Validation

- Release build
- node008 custom GPU suite — 10/10 passed
- focused boundary/tail, raw/compressible framing, suffix integrity, and legacy-stream ordering coverage
- `git diff --check`

Part of #292.

### DIFF
--- a/manual_server.cpp
+++ b/manual_server.cpp
@@ -3206,58 +3206,299 @@ int handle_manual_cuMemcpyHtoDAsync_v2(conn_t *conn) {
   return 0;
 }
 
-int handle_manual_cuMemcpyDtoH_v2(conn_t *conn) {
-  CUdeviceptr srcDevice = 0;
-  size_t byteCount = 0;
-  int request_id = 0;
+static int lupine_write_dtoh_chunk_response(conn_t *conn, int request_id,
+                                            CUresult result, const void *data,
+                                            size_t bytes) {
+  if (rpc_write_start_response(conn, request_id) < 0 ||
+      rpc_write(conn, &result, sizeof(result)) < 0 ||
+      (result == CUDA_SUCCESS && bytes != 0 &&
+       rpc_write_payload(conn, data, bytes) < 0) ||
+      rpc_write_end(conn) < 0) {
+    return -1;
+  }
+  return 0;
+}
+
+static int lupine_copy_dtoh_serial(conn_t *conn, int request_id,
+                                   CUdeviceptr src_device, size_t byte_count,
+                                   size_t offset) {
+  if (offset > byte_count) {
+    return -1;
+  }
+
   CUresult result = CUDA_ERROR_INVALID_VALUE;
-  std::vector<unsigned char> dstHost;
-
-  if (rpc_read(conn, &srcDevice, sizeof(srcDevice)) < 0 ||
-      rpc_read(conn, &byteCount, sizeof(byteCount)) < 0) {
-    return -1;
-  }
-
-  request_id = rpc_read_end(conn);
-  if (request_id < 0) {
-    return -1;
-  }
-
   size_t staging_size =
-      std::min(byteCount, (size_t)LUPINE_COMPRESS_BLOCK_BYTES);
+      std::min(byte_count - offset, (size_t)LUPINE_COMPRESS_BLOCK_BYTES);
+  std::vector<unsigned char> dst_host;
   if (staging_size != 0) {
     try {
-      dstHost.resize(staging_size);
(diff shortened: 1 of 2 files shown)


kevmo314:

PR #333, opened 2026-07-10
Pipeline async HtoD through a bounded pinned ring

### LINKED ISSUE #323: Performance: use a pinned-ring fast path for async HtoD staging
Part of #292. Highest-priority transfer-pipeline follow-up.

## Evidence

On node006, 256 MiB async HtoD plus synchronization takes 185.3 ms (1.349 GiB/s), versus 85.4 ms (2.929 GiB/s) natively. handle_manual_cuMemcpyHtoDAsync_v2 pins the entire byteCount before reading in chunks; pinning 256 MiB alone costs about 95 ms. #320 fixes cleanup but not allocation size.

## Critical async constraint

A strict two-slot ring that waits for an event before slot reuse can deadlock. Valid user code can enqueue a stream wait on an event, call cuMemcpyHtoDAsync with a payload larger than the ring, and record the releasing event only after the API returns. Full staging can enqueue and return; a blocking ring cannot.

## Safe first prototype

- Use two or three retained 4–8 MiB pinned slots as a nonblocking fast path.
- Read/decompress into a free slot, enqueue on the caller's stream, and record a disable-timing event.
- Before reuse, use cuEventQuery rather than an unconditional wait.
- If no slot has completed, fall back to the existing full-payload staging for the unread remainder so the async API can return.
- Keep capture handling on capture-owned storage.
- Reuse slots across calls/streams only after completion; clean up outside CUDA host callbacks.

Strict worst-case boundedness requires a larger protocol/ownership change and is not part of this first optimization.

## Required correctness coverage

- a pinned source whose stream waits on an event recorded only after cuMemcpyHtoDAsync returns (payload larger than ring)
- legacy default and explicit streams; sequential calls and cross-stream reuse
- immediate synchronization and dependent operations
- mid-payload RPC/CUDA errors with payload draining
- capture path unchanged; disconnect with copies in flight

## Measurement and acceptance

- benchmark 1–1024 MiB, random and compressible
- report ring hit/fallback rates, allocation time, event queries/waits, CPU, RSS, and pinned bytes
- compare slot count and 4/8 MiB sizes
- target at least 80% native loopback throughput on the fast path
- preserve async return behavior and avoid slow-link regression

kevmo314:
## Summary

- pipeline non-capture async HtoD payloads through a per-connection ring of four portable 8 MiB pinned slots on the POSIX fork-per-client server
- reclaim slots with disable-timing events and a bounded 2 ms query budget; if no slot completes, stage the unread suffix in one event-owned spill so the caller's stream is never synchronized
- serialize staging ownership across RPC lanes and invalidate context-owned buffers safely across create/attach/detach/destroy and primary-context release/reset, including same-handle reuse and deferred CUDA errors
- retain the full-request deferred-free path on Windows, where connections share one CUDA process and a safe ring would require process-wide context generations
- preserve CUDA graph capture-owned staging and fix response-lane ownership for concurrent RPC workers

This is stacked on #320. The performance commit is `fcaf3cd1`; once #320 lands, this PR's diff reduces to the ring implementation. Part of #292.

## Performance

Controlled on `inferable-node-008`, physical GPU 0 (RTX 2070 SUPER) selected with `CUDA_DEVICE_ORDER=PCI_BUS_ID CUDA_VISIBLE_DEVICES=0`, 256 MiB incompressible/pageable async HtoD followed by stream synchronization, 5 warmups and 15 measured copies per run:

|---|---:|---:|
| branch base | 217.325 ms | 1.150 GiB/s |
| this PR | 55.337 ms | 4.518 GiB/s |
| native CUDA | 20.710 ms | 12.072 GiB/s |

This is 3.93x baseline throughput and 74.5% lower latency, moving from 9.5% to 37.4% of native. Candidate repeats were 4.518, 4.413, and 4.544 GiB/s.

Across three candidate runs, the forked server cgroup used 3004.874 ms aggregate CPU and peaked at 52.3 MiB. Each connection retained two slots (16 MiB), recorded 640 submissions / 638 reclaims, and had zero poll exhaustion or spills. A slow-link repeat remained network-bound and spill-free at 2285.520 ms / 0.109 GiB/s.

## Correctness and validation

- `cmake --build build -j4 && ctest --test-dir build --output-on-failure`: 2/2 passed
- remote GPU custom suite: 12/12 passed on the post-format build
- gated pinned-source copy larger than the ring returns before its stream gate opens, then preserves the RPC payload snapshot
- cross-stream, legacy-stream, sequential reuse, concurrent RPC lane, explicit-context lifecycle, primary-context non-final release, reset, and same-raw-handle recreation tests pass
- focused legacy/direct/per-thread CUDA graph capture checks pass
- generated sources reproduce exactly; `git diff --check` and clang-format checks pass
- independent final concurrency/lifecycle/error-path review found no blocking issue

## Bounds and platform behavior

The retained Linux fast path is capped at 32 MiB per connection. A deliberately blocked stream can still allocate one full unread-suffix spill after the bounded poll; ambiguous completion failures remain quarantined until context teardown. Strictly bounded outstanding async DtoH/HtoD ownership needs a larger protocol change.

Windows deliberately keeps #320's full-request staging behavior because independent connections share primary contexts in one process; using this retained ring there without a process-wide generation registry would permit cross-client stale-handle reuse.

Closes #323.

### DIFF
--- a/codegen/codegen.py
+++ b/codegen/codegen.py
@@ -1460,6 +1460,52 @@ def write_client_post_call(f, function: Function, metadata: FunctionAnnotationMe
         f.write("    if (return_value == CUDA_SUCCESS && pKernel != nullptr) lupine_record_library_kernel(*pKernel, library, name, route);\n")
 
 
+SERVER_LIFECYCLE_FUNCTIONS = {
+    "cuDevicePrimaryCtxRetain",
+    "cuDevicePrimaryCtxRelease_v2",
+    "cuDevicePrimaryCtxReset_v2",
+    "cuCtxAttach",
+    "cuCtxDestroy_v2",
+    "cuCtxDetach",
+}
+
+
+def write_server_pre_call(f, function: Function):
+    name = function.name.format()
+    if name in SERVER_LIFECYCLE_FUNCTIONS:
+        f.write("    lupine_server_begin_lifecycle_transaction(conn);\n")
+    if name in {"cuDevicePrimaryCtxRelease_v2", "cuDevicePrimaryCtxReset_v2"}:
+        f.write("    lupine_server_prepare_primary_context(conn, dev);\n")
+    if name in {"cuCtxDestroy_v2", "cuCtxDetach"}:
+        f.write("    lupine_server_prepare_context_destroy(conn, ctx);\n")
+
+
+def write_server_post_call(f, function: Function):
+    name = function.name.format()
+    if name == "cuDevicePrimaryCtxRetain":
+        f.write(
+            "    lupine_server_note_primary_context(conn, dev, pctx, lupine_intercept_result);\n"
+        )
+    if name == "cuCtxDestroy_v2":
+        f.write(
+            "    lupine_server_finish_context_destroy(conn, ctx, lupine_intercept_result);\n"
+        )
+    if name == "cuCtxDetach":
+        f.write(
+            "    lupine_server_finish_context_detach(conn, ctx, lupine_intercept_result);\n"
+        )
+    if name == "cuDevicePrimaryCtxRelease_v2":
+        f.write(
+            "    lupine_server_finish_primary_context(conn, dev, false, lupine_intercept_result);\n"
+        )
+    if name == "cuDevicePrimaryCtxReset_v2":
+        f.write(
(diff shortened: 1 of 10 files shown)


kevmo314:

PR #362, opened 2026-07-17
Fix host callback dispatch framing

kevmo314:
## Summary

- replace raw wire callback pointers with client-owned callback tokens and validate them before invocation
- abort callback processing after failed transfer decode/allocation; drain only bounded, safely framed payloads and poison malformed frames
- add regression coverage for forced allocation failure, malformed frames, forged tokens, and next-message behavior

## Root cause

A transfer-loop failure only broke the loop, leaving the RPC reader inside the callback frame. The subsequent callback reads could therefore decode payload bytes as a function pointer.

## Validation

- `cmake --build build --target h2_test --parallel 8`
- `ctest --test-dir build --output-on-failure -R ^'h2_test$'`\n\nGPU-host note: inferable-node-008 exposes an RTX 2070 SUPER and RTX 4090 but does not have `nvcc`/the CUDA toolkit installed, so it cannot build the project there.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -58,6 +58,7 @@ set(CLIENT_SOURCES
     ${CMAKE_CURRENT_SOURCE_DIR}/compress.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/cache.cpp
     ${CMAKE_CURRENT_SOURCE_DIR}/checkpoint.cpp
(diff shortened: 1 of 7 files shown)


kevmo314:
Closing in favor of a minimal fix that closes the connection on callback-transfer failure; the token-registry and recovery logic are out of scope.

PR #364, opened 2026-07-17
Close callback connection on transfer failure

kevmo314:
## Summary

Close the RPC connection and stop its callback dispatch thread when a callback transfer cannot be read, allocated, or decoded.

## Root cause

The old `break` statements left the callback message partially unread, then interpreted the remaining payload as callback and user-data fields.

## Validation

- `cmake -S . -B build -DCMAKE_BUILD_TYPE=RelWithDebInfo`
- `cmake --build build --target lupine_driver h2_test --parallel 8`
- `ctest --test-dir build --output-on-failure -R ^'h2_test$'`

### DIFF
--- a/client.cpp
+++ b/client.cpp
@@ -7119,7 +7119,11 @@ void *rpc_client_dispatch_thread(void *arg) {
 
       int found = 0;
 
-      rpc_read(conn, &found, sizeof(int));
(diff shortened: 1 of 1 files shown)


kevmo314:
Closing: this only reproduces with deliberate client-side allocation fault injection, not normal usage. We should not add defensive recovery/teardown for an unobserved normal-path failure.

PR #369, opened 2026-07-24
test: build libNVVM samples from their group root instead of skipping them

### LINKED ISSUE #251: run_cuda_samples.sh: dsl/simple/ptxgen libNVVM entries never build (always SKIP:missing)
## Summary
Three `CORE_SAMPLES` entries in `test/run_cuda_samples.sh` — `dsl`, `simple`, `ptxgen` — never build and are silently reported as `SKIP:missing` on every run. They are the `7_libNVVM` compiler samples.

## Root cause (two compounding issues)
1. **Directory vs. executable name mismatch.** The sample directory is `Samples/7_libNVVM/device-side-launch`, but its CMake target/executable is `dsl` (`add_executable(dsl dsl.c)`). The runner lists `dsl`, but `resolve_sample_srcdir("dsl")` looks for a directory named `dsl` (none exists), so it short-circuits to `SKIP:missing`. Conversely, building `device-side-launch` would fail because the runner runs `cmake --build --target <sample>` and the target is `dsl`, not `device-side-launch`.
2. **Standalone configure lacks `NVVM_LIB`.** `dsl`/`simple`/`ptxgen` link `${NVVM_LIB}` / `${CUDA_LIB}`, which is only resolved by the parent `Samples/7_libNVVM/CMakeLists.txt`. The runner's per-sample `cmake -S <sample_dir>` does not load it, so even `simple`/`ptxgen` (whose directory names do match) cannot link.

## Options
- (cheap) Remove `dsl`, `simple`, `ptxgen` from `CORE_SAMPLES` since they are inert today.
- (proper) Teach the runner to configure the whole `Samples/7_libNVVM` tree and map directory names to their executable targets, then verify these libNVVM/device-side-launch samples actually work through the shim (they exercise `cuLaunchKernel` / device-side launch / libNVVM compilation and may hit unimplemented APIs).

mesutoezdil:
Closes #251

simple and ptxgen sit under Samples/7_libNVVM. Their CMake targets
link NVVM_LIB/CUDA_LIB, which only the parent 7_libNVVM CMakeLists.txt
resolves. The per-sample configure this runner uses did not load
that parent file, so both entries could never link and always
reported SKIP:missing or build-failed.

Rather than dropping them from CORE_SAMPLES, this teaches the runner
to route any sample belonging to that group through its group root
for configure/build/install, on both the modern CMake path and the
legacy Makefile path. Both routes land their binaries in a shared
install/bin alongside the runtime .ll/data files each sample needs.

Verified directly against the pinned nvidia/cuda:12.4.1-devel and
13.0.2-devel images: libnvvm is present and links cleanly in both;
the previous failures were purely a directory/build-routing bug in
this script, not a libnvvm availability problem. simple and ptxgen
are restored to CORE_SAMPLES now that they build and resolve
correctly.

### DIFF
--- a/test/run_cuda_samples.sh
+++ b/test/run_cuda_samples.sh
@@ -58,7 +58,7 @@ CORE_SAMPLES=(
   quasirandomGenerator_nvrtc
   simpleCudaGraphs streamOrderedAllocation cudaCompressibleMemory simpleZeroCopy alignedTypes LargeKernelParameter UnifiedMemoryPerf
   vectorAddMMAP
-  simpleHyperQ simpleVoteIntrinsics simpleAWBarrier binaryPartitionCG
+  simple simpleHyperQ simpleVoteIntrinsics simpleAWBarrier binaryPartitionCG
   globalToShmemAsyncCopy shfl_scan threadFenceReduction warpAggregatedAtomicsCG
   cdpSimplePrint cdpSimpleQuicksort cdpAdvancedQuicksort cdpQuadtree cdpBezierTessellation
   newdelete
@@ -75,7 +75,7 @@ CORE_SAMPLES=(
   imageDenoising marchingCubes particles simpleGL smokeParticles
   simpleTexture3D volumeFiltering volumeRender
   radixSortThrust segmentationTreeThrust template interval
-  ptxjit matrixMulDynlinkJIT threadMigration
+  ptxgen ptxjit matrixMulDynlinkJIT threadMigration
 )
 
 LIBRARY_SAMPLES=(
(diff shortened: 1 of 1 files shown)


kevmo314:
Is it possible to get libnvvm to work? Admittedly I haven't looked into the requirements of the library.

mesutoezdil:
> Is it possible to get libnvvm to work? Admittedly I haven't looked into the requirements of the library.

yeah it works. tested on both the 12.4 and 13.0 cuda images, libnvvm itself is fine, the bug was just that this script built simple/ptxgen alone instead of via their parent 7_libNVVM dir, soNVVM_LIB never got set.
and fixed! now builds through the group root, installs to a shared install/bin. Ran it for real..
ptxgen builds and outputs real PTX on both CUDA versions, simple builds and reaches the real driver call. Back in CORE_SAMPLES. hope it is what you wanna see

kevmo314:
> > Is it possible to get libnvvm to work? Admittedly I haven't looked into the requirements of the library.
> 
> 
> 
> yeah it works. tested on both the 12.4 and 13.0 cuda images, libnvvm itself is fine, the bug was just that this script built simple/ptxgen alone instead of via their parent 7_libNVVM dir, soNVVM_LIB never got set.
> 
> and fixed! now builds through the group root, installs to a shared install/bin. Ran it for real..
> 
> ptxgen builds and outputs real PTX on both CUDA versions, simple builds and reaches the real driver call. Back in CORE_SAMPLES. hope it is what you wanna see

Thanks, happy to see it works!

PR #370, opened 2026-07-24
sidecar: publish worker image to ghcr and use full ref

### LINKED ISSUE #287: Sidecar default image is unqualified and may not exist in a registry
## Problem

The sidecar's default container image reference is unqualified/unpublished. Fresh users cannot reliably resolve it, and resolution can vary by container runtime or registry search configuration.

## Proposed fix

Publish a versioned image to the project's chosen registry and use a fully qualified, immutable-or-versioned default. Document how releases update it and how users override it.

## Acceptance criteria

- The default image can be pulled anonymously in a clean environment.
- The reference includes an explicit registry/repository and version (or digest).
- CI validates the image build and a basic sidecar startup.
- Override behavior remains documented and tested.

mesutoezdil:
closes #287

`DEFAULT_IMAGE` in `python/lupine/sidecar.py` was an unqualified name
(`lupine-pytorch-worker:cuda-13.1.0`), so a fresh host w/ no local
build of that image couldn't resolve or pull it. This adds a workflow
that builds `deploy/Dockerfile.pytorch-worker` and pushes it to
`ghcr.io/<owner>/lupine-pytorch-worker:cuda-13.1.0`, and points
`DEFAULT_IMAGE` at the full ref.

Also bundled: an unrelated CI fix found while testing the above. The
main Dockerfile's `NVIDIA_UTILS_PACKAGE` pin (`nvidia-utils-535`)
recently became an empty transitional package on Ubuntu 24.04, so
every CUDA 12.6+ client image build failed to find `nvidia-smi`. The
nvidia-utils stage now falls back to whichever `nvidia-utils-NNN`,
newest first, actually ships the binary. This was breaking client
image builds on every open PR, not just this one.

## history

An earlier revision of this PR also tried to auto-select a worker
image matching the server's GPU driver by publishing four CUDA
versions and probing them at connect time. kevmo314 pointed out
(https://github.com/lupinemachines/lupine/pull/370#discussion_r3646194085)
that the probe (the existing ping/`cuda_available` handshake) isn't
real version negotiation: a client/worker pairing that's actually
incompatible can still pass the handshake and fail with random errors
later at runtime instead of failing the connection up front. Scoped
this PR back down to the original single-image change; the
auto-select/negotiation work is tracked separately in #374.

verified locally:
- built `deploy/Dockerfile.pytorch-worker` and confirmed it builds and
  publishes as `ghcr.io/<owner>/lupine-pytorch-worker:cuda-13.1.0`.
- built the main Dockerfile's nvidia-utils stage for CUDA 12.6.2: the
  535 pin downloads but has no `nvidia-smi`, falls back to
  `nvidia-utils-610`, and the resulting `/nvidia-smi` is a valid
  stripped x86-64 ELF binary. Rebuilt it for CUDA 12.4.1 (previously
  passing) to confirm no regression there.

### DIFF
--- a/.github/workflows/publish-pytorch-worker-image.yml
+++ b/.github/workflows/publish-pytorch-worker-image.yml
@@ -1,0 +1,63 @@
+name: Publish PyTorch Worker Image
+
+on:
+  release:
+    types: [published]
+  push:
+    branches:
+      - main
+    paths:
+      - deploy/Dockerfile.pytorch-worker
+      - .github/workflows/publish-pytorch-worker-image.yml
+  workflow_dispatch:
+
+permissions:
+  contents: read
+  packages: write
+
+jobs:
+  build:
+    name: PyTorch worker image
+    runs-on: ubuntu-latest
+    timeout-minutes: 45
+    steps:
+      - name: Check out repository
+        uses: actions/checkout@v6
+
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v3
+
+      - name: Log in to GHCR
+        uses: docker/login-action@v3
+        with:
+          registry: ghcr.io
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Prepare image tags
+        id: image-tags
+        run: |
+          set -euo pipefail
+          {
+            echo 'tags<<EOF'
+            echo 'ghcr.io/${{ github.repository_owner }}/lupine-pytorch-worker:cuda-13.1.0'
+            if [ '${{ github.event_name }}' = 'release' ]; then
+              echo 'ghcr.io/${{ github.repository_owner }}/lupine-pytorch-worker:${{ github.event.release.tag_name }}'
+            fi
+            echo 'EOF'
+          } >> "$GITHUB_OUTPUT"
+
+      - name: Build and publish worker image
+        uses: docker/build-push-action@v7
+        with:
+          context: .
+          file: deploy/Dockerfile.pytorch-worker
+          platforms: linux/amd64
+          push: true
+          build-args: |
+            CUDA_VERSION=13.1.0
+            UBUNTU_VERSION=24.04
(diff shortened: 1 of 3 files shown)


kevmo314 (approved):

.github/workflows/publish-pytorch-worker-image.yml
```diff
@@ -0,0 +1,63 @@
...
+
+      - name: Build and publish worker image
+        uses: docker/build-push-action@v7
+        with:
+          context: .
+          file: deploy/Dockerfile.pytorch-worker
+          platforms: linux/amd64
+          push: true
+          build-args: |
+            CUDA_VERSION=13.1.0
```
kevmo314, line 58: Hmm this is actually a little tricky because the version has to be <= the server version. So if the user does not run a server with 13.1 basically this sidecar is unusable. That being said it seems like a poor experience to require the user to choose a cuda version...

mesutoezdil: > Hmm this is actually a little tricky because the version has to be <= the server version. So if the user does not run a server with 13.1 basically this sidecar is unusable. That being said it seems like a poor experience to require the user to choose a cuda version...

hmm i try other way now.. sent you soon

mesutoezdil: The scope has become broader than it was at the start 
Is that suitable for you? I explained it in the description.

kevmo314: Hmm I don't think it works as written, the connection with a client that's too new will still succeed but you'll see random errors at runtime. We should maybe do some signaling in the connection to make the connection actually fail. In light of that point taken that the scope is too large. For this PR you can just publish the 13.1 image (your original code) and we can follow up with the version negotiation code separately. If you'd like to do it that would be excellent but feel free to file an issue and I can pick it up in the next few days too.

mesutoezdil: sure, i did

PR #373, opened 2026-07-24
add Session.device to match SidecarSession

mesutoezdil:
python/examples/tensor.py calls session.device() right after connect().
SidecarSession (python/lupine/sidecar.py) defines device(index=0), but
the native Session in python/lupine/__init__.py, returned by connect()
on any non-macOS-fallback path, only has devices() (plural). Running
this example against a native LUPINE server raises
AttributeError: 'Session' object has no attribute 'device' at line 24,
before the script does anything useful.

Add Session.device(index=0) that returns self.devices()[index],
mirroring SidecarSession's existing method, so the example works the
same way regardless of which session type connect() returns.

Added a direct test for the new method.


python/lupine/__init__.py
```diff
@@ -140,6 +140,11 @@ def devices(self) -> list[Any]:
         count = int(torch.cuda.device_count())
         return [torch.device("cuda", index) for index in range(count)]
 
+    def device(self, index: int = 0) -> Any:
+        """Return one GPU from LUPINE's native virtual device topology."""
+
+        return self.devices()[index]
```
kevmo314:
```suggestion
        torch = _torch()
        if count >= torch.cuda.device_count():
          raise IndexError()
        return torch.device("cuda", index)
```

mesutoezdil: count isn't defined here, would NameError. also drops negative index support. fixed in 302bc6f with range(count)[index]

kevmo314 (approved):

PR #375, opened 2026-07-25
server: fix count times sizeof overflow in generated transfer sizing

### LINKED ISSUE #249: server: peer-controlled count overflows count*sizeof(elem) in generated transfer sizing (OOB read)
## Summary
Generated transfer sizing multiplies a peer-controlled count by an element size without an overflow check, which can be exploited to make the server child `malloc`/`read` a small buffer while passing the original (huge) count into the CUDA call, causing an out-of-bounds read and a server-child crash.

## Location
`codegen/ops.py` `transfer_size_expr()` (around line 162):

```python
def transfer_size_expr(self) -> str:
    if self.is_void_bytes:
        return self.byte_count_expr()
    return f"{self.element_count_expr()} * sizeof({self.ptr.ptr_to.format()})"
```

The emitted C for a variable-length array parameter is `{count} * sizeof({elem})`. `count` comes from the RPC (peer-controlled). When it wraps (e.g. `0x1000000000000001 * 16 == 16`), the wrapped size is `malloc`'d/`read`, but the **original** `count` is forwarded to the CUDA driver call.

## Repro
Against a running lupine server, via the client shim:
```c
CUgraph g1, g2; cuGraphCreate(&g1,0); cuGraphCreate(&g2,0);
CUgraphNode node, deps[1]={0};
size_t nd = 0x1000000000000001ULL;          // nd * 16 wraps to 16
cuGraphAddChildGraphNode(&node, g1, deps, nd, g2);  // server OOB-reads -> child dies
```
Observed: server child crashes, the RPC returns `CUDA_ERROR_DEVICE_UNAVAILABLE`.

## Suggested fix
Route the size through a checked multiply, e.g. add `lupine_checked_mul_size(count, sizeof(elem))` returning `SIZE_MAX` on overflow in `codegen/ops.py`, and rely on the existing `if (size != 0 && ptr == nullptr) goto ERROR` guard to reject the request. Regenerate `codegen/gen_{server,client}.cpp`.

## Status
Documented in `BUG_REPORT.md` (Finding #3). Not yet fixed — regeneration of the generated sources requires the CI codegen toolchain.

mesutoezdil:
Fixes #249
A peer sent count could overflow when multiplied by sizeof(elem) in
generated transfer sizing. The wrapped small size was used for the
malloc, but the real huge count was still passed to the CUDA call.
This caused an out of bounds read and a server crash.

Repro from the issue:
```c
size_t nd = 0x1000000000000001ULL;   // nd * 16 wraps to 16
cuGraphAddChildGraphNode(&node, g1, deps, nd, g2);
```

Add `lupine_checked_mul_size(a, b)` in `lupine_platform.h`. It
saturates to SIZE_MAX on overflow instead of wrapping, so malloc
fails and the existing null check rejects the request.

Route every count times sizeof site in `codegen/ops.py` through it:
`ArrayOperation.transfer_size_expr`, `OptionalArrayOperation` (3
sites), `DeepStructOperation` (4 sites, one of them a vector resize
that would otherwise throw on a huge size, so it short circuits
before resize instead).

Regenerated `gen_server.cpp`, `gen_client.cpp`, `gen_nvml_client.inc`,
`gen_nvml_server.inc` with real CUDA 13.1.0 headers. Every changed
hunk is only this fix, checked line by line.

Test plan

- `docker build --target server` and `--target client`: both build
  clean.
- Full CMake build plus `ctest`: 100% tests passed (6 passed, 2
  skipped for unrelated reasons, 0 failed).
- Added `lupine_platform_test.cpp` for the new helper, passes.
- Confirmed in `gen_server.cpp` that the issue repro now hits
  `malloc(SIZE_MAX)` -> null -> clean reject instead of an out of
  bounds read.
- Did not run the real network repro against a live GPU server, that
  needs real hardware. Should be covered by this repo's own GPU
  Integration CI on this PR.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -176,6 +176,12 @@ else()
     add_test(NAME checkpoint_test COMMAND checkpoint_test)
     set_tests_properties(checkpoint_test PROPERTIES TIMEOUT 10)
 
+    add_executable(lupine_platform_test
+        ${CMAKE_CURRENT_SOURCE_DIR}/lupine_platform_test.cpp
+    )
+    add_test(NAME lupine_platform_test COMMAND lupine_platform_test)
+    set_tests_properties(lupine_platform_test PROPERTIES TIMEOUT 10)
+
     add_library(test_lupinecr_provider SHARED
         ${CMAKE_CURRENT_SOURCE_DIR}/test/test_checkpoint_provider.c
     )
--- a/codegen/gen_client.cpp
+++ b/codegen/gen_client.cpp
@@ -152,14 +152,17 @@ CUresult cuDeviceGetName(char *name, int len, CUdevice dev) {
       rpc_write(conn, &len, sizeof(int)) < 0 ||
       rpc_write(conn, &dev, sizeof(CUdevice)) < 0 ||
       rpc_wait_for_response(conn) < 0 ||
-      (lupine_prepare_host_range_write(name, len * sizeof(char)), false) ||
-      (len * sizeof(char) != 0 &&
-       rpc_read(conn, name, len * sizeof(char)) < 0) ||
+      (lupine_prepare_host_range_write(
+           name, lupine_checked_mul_size(len, sizeof(char))),
+       false) ||
+      (lupine_checked_mul_size(len, sizeof(char)) != 0 &&
+       rpc_read(conn, name, lupine_checked_mul_size(len, sizeof(char))) < 0) ||
       rpc_read(conn, &return_value, sizeof(CUresult)) < 0 ||
       rpc_read_end(conn) < 0)
     return CUDA_ERROR_DEVICE_UNAVAILABLE;
   if (return_value == CUDA_SUCCESS)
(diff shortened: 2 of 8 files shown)


kevmo314:
> It saturates to SIZE_MAX on overflow instead of wrapping, so malloc fails and the existing null check rejects the request.

There's no real difference between crashing the server and rejecting the request. Practically, we cannot reject CUDA calls because it will change the semantic between running locally and via lupine.

This PR is more correct, but it adds ~300 lines of code for basically the same end result. Unless you're observing that this can cause _other_ connections to crash? In which case that would be worth fixing.

Btw, at a project philosophy level, the lupine client is generally treated as trusted. Since you can basically do whatever you want to the memory space with CUDA calls, we don't really need to be that defensive about crashing the server, validation, etc. It is simpler to blindly forward to/from CUDA.

mesutoezdil:
yeah checked this, y r right. server forks per connection so this only takes down the attacker's own child, nth else touches it.. 
not worth the extra code for basically the same outcome then. 
and closing this 1, thanks for pushing back on it

PR #390, opened 2026-07-27
ci: install torch extra so python tests actually run

### LINKED ISSUE #389: python CI never installs torch, sidecar and tensor tests silently skip
python/pyproject.toml only lists torch as an optional extra, not in the
dev group. Both python.yml and publish.yml run plain `uv run pytest`,
no `--extra torch` or `--all-extras`. sidecar_test.py and tensor_test.py
both start with `torch = pytest.importorskip("torch")`, so without torch
installed the whole file gets skipped instead of run.

Ran the exact CI command in a clean venv to check: `uv run pytest` gives
"20 passed, 2 skipped". Adding `--all-extras` gives "43 passed, 1
skipped". So 23 real tests (ContainerRuntime, SidecarSession, tensor
encode/decode, dispatch mode forwarding) never actually run in CI.

publish.yml has the same gap, so this also gates the actual PyPI
release, not just PR checks.

Fix

Add `--all-extras` (or `--extra torch`) to the `uv run pytest` step in
both python.yml and publish.yml.

mesutoezdil:
Fixes #389

torch is only listed as an optional extra in pyproject.toml, not in
the dev group. python.yml and publish.yml both ran plain `uv run
pytest`, so torch never got installed. sidecar_test.py and
tensor_test.py both start with `pytest.importorskip("torch")`, so
those files silently skipped instead of running.

Added `--all-extras` to the three affected `uv run pytest` steps
(python.yml's main run and its Python 3.9 compat run, plus
publish.yml).

Test plan
Ran the exact commands in a clean venv:
- without the extra: 20 passed, 2 skipped
- with `--all-extras`: 42 passed, 1 skipped
- same result on Python 3.9

So this brings 22 real tests back into CI, and fixes the same gap in
the actual PyPI publish workflow.


kevmo314 (approved):

kevmo314:
Don't worry about merging this change looks safe. I'll get it into main

PR #409, opened 2026-07-30
Declare hand-written client functions in generated code

kevmo314:


kevmo314:
Not worth the churn: this only automates the declaration for gen_client.cpp's functionMap when a CUDA-12+ function is marked @disabled client. It does not reduce cuda_compat.h maintenance (client.cpp is a separate TU and never sees generated declarations), the failure mode is a loud build error rather than a silent bug, and #407 already fixed the one occurrence.

PR #412, opened 2026-07-30
Keep the sidecar worker alive when a request cannot be serialized

mesutoezdil:
`_send` built the JSON header inside the transport `try` block, so an argument that JSON cannot encode killed the worker and reported a misleading transport failure. Nothing had been written to the pipe at that point, so the session was still usable.

Reproduced against the real worker over pipes:

```
complex fill: SidecarError cause=TypeError: Object of type complex is not JSON serializable
complex fill: worker killed=True, session dead=True
generator:    SidecarError cause=TypeError: Object of type Generator is not JSON serializable
generator:    worker killed=True, session dead=True
```

Both are reachable from the public API, `torch.full((2, 2), 1 + 2j, device=dev)` and `torch.randn(3, generator=g, device=dev)`, and each one took down the whole sidecar session.

Serialize the header before the transport block and raise a `SidecarError` that names the problem, leaving the worker running.

New test spawns the worker, sends a complex fill value, and pings afterwards to prove the session survives. It fails without the fix.


python/lupine/sidecar.py
```diff
@@ -313,6 +313,12 @@ def _send(
             request["tensor_streams"] = [metadata for _, metadata in input_tensors]
             try:
                 header = json.dumps(request).encode("utf-8") + b"\n"
+            except (TypeError, ValueError) as exc:
+                # Nothing reached the worker yet, so the session stays usable.
+                raise SidecarError(
+                    f"sidecar cannot serialize the request: {exc}"
```
kevmo314: Can we use pickle or something so it's guaranteed to be serializable? If a user hits this they will not know what to do next to resolve the issue or why it's problematic.

mesutoezdil: pickle would work, both are picklable. but the image's torch isn't the user's torch, and unpickling its response lets the image run code in the user's process. it wouldn't drop this branch either, pickle raises too.

message is fair though. now:

`sidecar cannot send this operation to the worker: Object of type complex is not JSON serializable; arguments must be tensors, numbers, strings, dtypes, devices, layouts, or memory formats. Build the value on CPU and move it with .to(device) instead.`

complex scalars are the only type users hit by accident. i can encode those in four lines, here or follow-up.

kevmo314 (approved):

PR #523, opened 2026-08-18
Add Prometheus GPU monitoring

### LINKED ISSUE #522: [Feature] Add Prometheus GPU monitoring and client attribution
## Problem

NVML reports the Lupine server child that owns a CUDA context, but it does not identify the remote client behind that process. Lupine needs Prometheus metrics for host GPU usage, per-client usage, and the client-to-server PID mapping.

### Client identity

After the HTTP/2 handshake, the server advertises whether it accepts client metadata. A supporting client sends only automatically available data:

```cpp
struct lupine_client_metadata {
  uint64_t client_pid;
  uint64_t client_pid_namespace_inode;
  char connection_kind[16];
  char client_process_name[128];
  char client_hostname[128];
};
```

The PID namespace inode distinguishes containers that both use PID 1. It is omitted outside Linux. No environment variables are required.

### Host PID discovery

The server stores active connections in a process-shared registry. Around the first API that creates or retains a CUDA context, it compares the NVML process list before and after the call:

```cpp
lupine_monitoring_begin_context_create(device);
CUresult result = cuCtxCreate_v2(&context, flags, device);
lupine_monitoring_end_context_create(result == CUDA_SUCCESS);
```

The PID is saved only when there is one new, unused candidate. Discovery is serialized to avoid assigning the same NVML PID to concurrent clients. The parent removes the exact registry entry when `waitpid` reaps a child.

### Metrics endpoint

Lupine already has one HTTP listener. Prior-knowledge HTTP/2 remains the RPC path, while HTTP/1.1 `GET /metrics` returns Prometheus text on the same port:

```bash
curl http://<server>:14833/metrics
```

NVML is sampled only when this endpoint is requested. Per-process utilization uses an independent two-second freshness window, so concurrent scrapes do not consume each other's samples and old busy samples expire after the process becomes idle. Failed NVML fields are omitted instead of being reported as valid zeroes.

The server resolves the real NVML functions through its existing optional runtime loader. This avoids a hard `libnvidia-ml.so.1` dependency when the unified server runs without NVIDIA NVML.

## Metrics

| Metric | Meaning |
|---|---|
| `lupine_monitor_nvml_up` | `1` when NVML initialized, otherwise `0`. |
| `lupine_monitor_process_utilization_supported` | Per-device flag showing whether NVML supports process SM utilization. |
| `lupine_monitor_dropped_connections_total` | Connections not tracked because the shared registry was full. |
| `lupine_monitor_host_pid_discovery_failures_total` | Context discoveries that did not resolve a host PID. |
| `lupine_monitor_host_pid_discovery_ambiguities_total` | Discoveries with multiple or already-used PID candidates. |
| `lupine_server_connection_info` | Active client PID, server child PID, host PID, process name, hostname, address, and connection type mapping. |
| `lupine_host_gpu_memory_total_bytes` | Physical GPU memory capacity. |
| `lupine_host_gpu_memory_used_bytes` | Physical GPU memory currently used. |
| `lupine_host_gpu_utilization_percent` | Physical GPU utilization from 0 to 100. |
| `lupine_client_device_memory_used_bytes` | GPU memory attributed to one remote client. |
| `lupine_client_device_utilization_percent` | SM utilization attributed to one remote client, from 0 to 100. |

Host metrics use device index, type, and UUID labels. Client metrics also include the generated client ID, address, hostname, and process name. A series is omitted when its NVML query is unavailable or fails.

## Validation

Release build with CUDA 13.0.88:

```text
100% tests passed, 0 tests failed out of 12
```

- Four concurrent clients mapped to four distinct host PIDs reported by `nvidia-smi`.
- Each client metric reported 335544320 bytes, matching `nvidia-smi` at 320 MiB.
- Three consecutive busy scrapes reported 100%, matching host utilization.
- A busy-to-idle test changed from 84% to 0% after the freshness window.
- Metrics disappeared after disconnect.
- A non-monitoring server build passed.
- `readelf` showed no `libnvidia-ml.so.1` runtime dependency.

maverick-woo:
## What changed

- Serve `GET /metrics` as HTTP/1.1 on the existing RPC port. HTTP/2 RPC traffic is unchanged.
- Collect host and per-client GPU memory and utilization from NVML when `/metrics` is requested.
- Map each remote client to its Lupine server child and NVML host PID.
- Negotiate optional client metadata during the HTTP/2 handshake, so clients skip it when the server does not support it.
- Load the real NVML library at runtime, keeping the unified server usable when NVML is unavailable.

There are no monitoring-specific environment variables, background threads, timers, or extra listening ports.

## Validation

- Release build with CUDA 13.0.88: 12/12 CTest tests passed.
- A build without NVML monitoring also passed.
- The server binary has no `DT_NEEDED` dependency on `libnvidia-ml.so.1`.
- Four concurrent clients produced four correct client-to-host PID mappings; each 320 MiB allocation matched `nvidia-smi` exactly.
- Three consecutive scrapes during a busy workload all reported 100%, matching `nvidia-smi`.
- A busy-to-idle workload changed from 84% to 0% after the fixed two-second sample window.
- Client metrics disappeared after disconnect.

Closes #522


copy_pipeline.cpp
```diff
@@ -1106,7 +1107,9 @@ int handle_manual_cuCtxCreate_v2(conn_t *conn) {
 
   CUcontext context = nullptr;
   lupine_server_begin_lifecycle_transaction(conn);
+  lupine_monitoring_begin_context_create(conn, device);
```
kevmo314: Why this API specifically?

monitor_client.cpp
```diff
@@ -0,0 +1,175 @@
...
+namespace {
+
+bool env_truthy(const char *value) {
+  return value != nullptr && value[0] != '\0' && strcmp(value, "0") != 0 &&
+         strcasecmp(value, "false") != 0 && strcasecmp(value, "no") != 0 &&
+         strcasecmp(value, "off") != 0;
+}
+
+bool metadata_enabled() {
+  const char *explicit_value = getenv("LUPINE_CLIENT_METADATA");
```
kevmo314: I agree this should gracefully degrade instead, instead of an env var.

monitor_client.cpp
```diff
@@ -0,0 +1,175 @@
...
+  }
+  dst[0] = '\0';
+  if (gethostname(dst, static_cast<int>(size)) != 0) {
+    dst[0] = '\0';
+  }
+  dst[size - 1] = '\0';
+}
+
+uint64_t process_start_time_ticks() {
+  std::ifstream input("/proc/self/stat");
```
kevmo314: This is Linux-specific, you'll need to add a Windows version too or omit the field on Windows.

monitor_client.cpp
```diff
@@ -0,0 +1,175 @@
...
+  std::string value;
+  if (!std::getline(input, value)) {
+    dst[0] = '\0';
+    return;
+  }
+  copy_value(dst, value.c_str());
+}
+
+template <size_t N> void fill_process_name(char (&dst)[N]) {
+  const char *override_name = getenv("LUPINE_CLIENT_NAME");
```
kevmo314: What's this for?

monitor_client.cpp
```diff
@@ -0,0 +1,175 @@
...
+  lupine_client_metadata_v2 metadata = {};
+  metadata.client_pid = process_id();
+  metadata.client_start_time_ticks = process_start_time_ticks();
+  metadata.client_pid_namespace_inode = pid_namespace_inode();
+  metadata.connection_index = connection_index;
+  copy_value(metadata.connection_kind, connection_kind);
+  fill_process_name(metadata.client_process_name);
+  copy_first_line(metadata.client_boot_id,
+                  "/proc/sys/kernel/random/boot_id");
+  copy_env(metadata.vdevice_index, "LUPINE_VDEVICE_INDEX");
```
kevmo314: I'm not sure lupine is the right place for data like this if it doesn't affect execution in any way. What's the motivation for it?

monitor_server.cpp
```diff
@@ -0,0 +1,1164 @@
...
+
+bool atomic_compare_exchange(uint32_t *value, uint32_t expected,
+                             uint32_t desired) {
+  return __atomic_compare_exchange_n(value, &expected, desired, false,
+                                     __ATOMIC_ACQ_REL, __ATOMIC_ACQUIRE);
+}
+
+void atomic_increment(uint64_t *value) {
+  __atomic_fetch_add(value, 1, __ATOMIC_RELAXED);
+}
```
kevmo314: Inline these functions

monitor_server.cpp
```diff
@@ -0,0 +1,1164 @@
...
+    source = &reinterpret_cast<const sockaddr_in6 *>(&address)->sin6_addr;
+  }
+  if (source == nullptr ||
+      inet_ntop(address.ss_family, source, text, sizeof(text)) == nullptr) {
+    return "unknown";
+  }
+  return text;
+}
+
+struct nvml_api {
```
kevmo314: Why does nvml need to be stubbed here? You can call the real nvml api as it's not shimmed on the server.

monitor_server.cpp
```diff
@@ -0,0 +1,1164 @@
...
+  nvml_api *active_api = api.open() ? &api : nullptr;
+  std::vector<unsigned long long> last_timestamps;
+  unsigned int interval = metrics_interval_ms();
+  std::string metrics = render_metrics(active_api, &last_timestamps);
+  auto next_sample = std::chrono::steady_clock::now() +
+                     std::chrono::milliseconds(interval);
+  for (;;) {
+    auto now = std::chrono::steady_clock::now();
+    if (now >= next_sample) {
+      metrics = render_metrics(active_api, &last_timestamps);
```
kevmo314: Isn't this rather expensive to run on a cron? Wouldn't it be better to run it on request?

monitor_server.cpp
```diff
@@ -0,0 +1,1164 @@
...
+  if (errno != 0 || end == text || *end != '\0' || parsed == 0 ||
+      parsed > 65535) {
+    return false;
+  }
+  *value = static_cast<uint16_t>(parsed);
+  return true;
+}
+
+unsigned int metrics_interval_ms() {
+  const char *text = getenv("LUPINE_METRICS_INTERVAL_MS");
```
kevmo314: Let's just hardcode this to 1000ms.

monitor_server.cpp
```diff
@@ -0,0 +1,1164 @@
...
+      __atomic_thread_fence(__ATOMIC_ACQUIRE);
+      uint32_t after = atomic_load(&slot->sequence);
+      if (before == after && (after & 1U) == 0 &&
+          atomic_load(&slot->state) == kSlotActive) {
+        copied = true;
+        break;
+      }
+    }
+    if (copied && snapshot.server_pid > 0) {
+      if (kill(snapshot.server_pid, 0) != 0 && errno == ESRCH) {
```
mesutoezdil: if a child dies uncleanly this slot only gets freed when this check happens to fire, and if the pid gets reused first it sticks around attributing usage to the wrong client, maybe simpler to have the parent free the slot on waitpid reap and drop this sweep entirely?

monitor_server.cpp
```diff
@@ -0,0 +1,1164 @@
...
+}
+
+void serve_request(int listener, const std::string &metrics) {
+  int client = accept(listener, nullptr, nullptr);
+  if (client < 0) {
+    return;
+  }
+  timeval timeout = {};
+  timeout.tv_sec = 2;
+  setsockopt(client, SOL_SOCKET, SO_RCVTIMEO, &timeout, sizeof(timeout));
```
mesutoezdil: there's a rcv timeout here but send_all has no snd one, so a scraper that stops reading blocks send() forever and wedges the whole loop, prob worth a SO_SNDTIMEO too

monitor_server.cpp
```diff
@@ -0,0 +1,1164 @@
...
+          NVML_SUCCESS ||
+      probe->device == nullptr) {
+    note_discovery_failure();
+    return;
+  }
+  if (!lock_host_pid_discovery()) {
+    note_discovery_failure();
+    return;
+  }
+  probe->lock_held = true;
```
mesutoezdil, line 365: iiuc this cross-process lock is held across cuCtxCreate plus up to ~1s of nvml polling, so every client's first ctx create serializes, might be fine but sth to keep in mind on a busy server w/ many concurrent connects

copy_pipeline.cpp
maverick-woo: This for lookup hostpid if server in container

monitor_client.cpp
maverick-woo: Agreed. I will remove the environment variable and negotiate this capability during the connection handshake. The server will advertise client metadata support, and the client will only send metadata when that capability is present, so connections to older servers degrade gracefully.

monitor_client.cpp
maverick-woo: Agreed. I’ll guard the Linux /proc fields with _WIN32 and omit them on Windows for now. They will be reported as zero or empty, while the process PID and other available metadata remain populated.

monitor_client.cpp
maverick-woo: support one client use multiprocesses

monitor_client.cpp
maverick-woo: This can be delete

monitor_server.cpp
maverick-woo: agree

monitor_server.cpp
maverick-woo: agree, this can call directly

monitor_server.cpp
maverick-woo: Agreed. I will switch this to request-triggered collection with a 1-second cache: `/metrics` will refresh the metrics only when the cache is older than one second; otherwise it will return the cached result. This avoids NVML work when there are no scrapes and limits collection to at most once per second across multiple scrapers. I will also move stale child-slot cleanup to the `waitpid` reap path instead of relying on metrics rendering.

monitor_server.cpp
maverick-woo: ok

CMakeLists.txt
```diff
@@ -343,6 +344,12 @@ if(NOT WIN32 AND LUPINE_BUILD_HIP)
 endif()
 
 if(LUPINE_BUILD_SERVER)
+    set(LUPINE_MONITORING_ENABLED OFF)
```
kevmo314: We can set this to ON all the time, we don't need a switch to toggle it off.

monitor_server.cpp
```diff
@@ -0,0 +1,926 @@
...
+    return;
+  }
+  std::unique_ptr<host_pid_probe> probe(new (std::nothrow) host_pid_probe());
+  if (probe == nullptr || !probe->session.active()) {
+    note_discovery_failure();
+    return;
+  }
+  char pci_bus_id[64] = {};
+  auto get_device = LUPINE_NVML_SYMBOL(nvmlDeviceGetHandleByPciBusId);
+  if (cuDeviceGetPCIBusId(pci_bus_id, sizeof(pci_bus_id), cuda_device) !=
```
kevmo314, line 353: We might need a lock here to disable the metrics loop when cuda graph capture is going on. During cuda graph capture these calls might get captured unintentionally?

monitor_client.cpp
maverick-woo: I removed process start time entirely. The PID namespace inode is now an optional Linux-only value used to distinguish containers that both use PID 1; Windows omits it and uses GetModuleFileNameA for the process name.

monitor_client.cpp
maverick-woo: The override has been removed. The client now reports only the process name detected from its executable, with no environment variable.

monitor_client.cpp
maverick-woo: Implemented. The server advertises a capability during the HTTP/2 handshake, and the client sends metadata only when that bit is present. Tests cover supported, absent, and rejected metadata while the next RPC still continues.

copy_pipeline.cpp
maverick-woo: cuInit does not create a CUDA context, so the process is not visible in NVML yet. The new code probes only around cuCtxCreate_v2 and cuDevicePrimaryCtxRetain, which are the first points where an NVML PID can appear.

monitor_server.cpp
maverick-woo: Implemented on demand. There is no monitor process, polling thread, timer, or response cache. Each GET /metrics performs one collection. The fixed two-second NVML lastSeen window only defines sample freshness and lets concurrent scrapes read the same recent samples.

monitor_client.cpp
maverick-woo: Removed. Workload ID, vdevice index, boot ID, connection index, and their environment variables are no longer part of the protocol. Metadata is limited to fields needed for process attribution.

monitor_server.cpp
maverick-woo: The custom NVML API table is gone. Monitoring calls the real NVML functions resolved through the same optional runtime loader used by the NVML server backend. This keeps the unified server from gaining a hard libnvidia-ml dependency; readelf shows no NVML DT_NEEDED entry.

monitor_server.cpp
maverick-woo: The atomic wrappers and seqlock are gone. Short registry operations now use one robust process-shared mutex, while host PID discovery keeps a separate mutex.

monitor_server.cpp
maverick-woo: The periodic sampler and its interval have been removed, so there is no timer or interval setting to configure.

monitor_server.cpp
maverick-woo: I kept serialization because overlapping before/after NVML process-list differences cannot be attributed safely. It runs only until a connection resolves its first host PID; later context calls skip it. A four-client concurrent startup test produced four distinct mappings that all matched nvidia-smi. Per-device locks can be a follow-up if startup contention becomes material.

monitor_server.cpp
maverick-woo: Implemented. The HTTP/1.1 path sets both SO_RCVTIMEO and SO_SNDTIMEO to two seconds before reading or writing the request.

monitor_server.cpp
maverick-woo: Implemented. The parent now frees the exact registry slot when waitpid reaps the connection child. The PID liveness sweep has been removed.

CMakeLists.txt
maverick-woo: ok

monitor_server.cpp
maverick-woo: This should not interfere with CUDA Graph capture. `/metrics` runs in a separate process and only queries NVML. `cuDeviceGetPCIBusId()` is called once during context creation, before graph capture can start, and it does not enqueue GPU work. So an extra lock should not be needed here.

CMakeLists.txt
maverick-woo: Done. I removed the internal toggle, so monitoring is always built for supported Linux CUDA+NVML server builds. There is no user-facing switch. The Release build and all 12 tests pass.

monitor_server.cpp
```diff
@@ -0,0 +1,926 @@
...
+
+int handle_lupine_client_metadata(conn_t *conn) {
+  lupine_client_metadata_header header = {};
+  if (conn == nullptr ||
+      rpc_read(conn, &header, sizeof(header)) != sizeof(header)) {
+    return -1;
+  }
+  int status = 0;
+  lupine_client_metadata metadata = {};
+  if (header.payload_size > LUPINE_CLIENT_METADATA_MAX_PAYLOAD) {
```
mesutoezdil, line 702: returning -1 here kills the connection on an oversized payload, so a future client with a bigger metadata struct gets disconnected by older servers. the previous revision drained it and answered with a status, seems safer to keep.

maverick-woo: Good catch. This was a regression during the simplification. I restored draining the oversized payload and returning status 2, so older servers reject the metadata without dropping the RPC connection. Release build and all 12 tests pass.

PR #660, opened 2026-08-27
Serve HTTP/1.x requests on the RPC port

mesutoezdil:
Follows the single-port discussion in #523.

## What changed

The connection child now peeks at the first bytes of an accepted connection before any RPC state exists:

- the fixed 24-byte HTTP/2 client preface proceeds into the existing nghttp2 path, with nothing consumed and `h2.cpp` untouched
- anything else is answered as a single HTTP/1.x request: `HEAD /` and `GET /` mirror the HTTP/2 version probe (including `x-lupine-cuda-version`), every other path returns 404
- undecidable connections time out after 5s

The RPC/non-RPC split stays a routing decision rather than an HTTP version one: the peek only detects framing, and both layers route by path (`POST /` is RPC). HTTP/1.1-only consumers such as Prometheus get answered on the same port, an HTTP/2 client asking for `GET /metrics` would land on the same route once it exists, and an HTTP/3 listener can later feed the same seam via Alt-Svc.

`/metrics` itself is intentionally not here; #523 can drop its separate `LUPINE_METRICS_PORT` listener and register the endpoint on this seam instead.

## Validation

- new `dispatch_test` covers: full/split preface reaching the RPC path with nothing consumed, `GET /` and `HEAD /` responses, 404 for unknown paths, closed-peer failure
- `dispatch_test`, `transport_test`, and `h2_test` pass locally (core-only build, macOS)
- GPU integration is CI-only for me; happy to iterate if the L4 jobs catch anything

### DIFF
--- a/dispatch.cpp
+++ b/dispatch.cpp
@@ -1,0 +1,186 @@
+// Protocol dispatch for the server's single listening port. Lupine clients
+// speak prior-knowledge HTTP/2 and always open with the fixed 24-byte client
+// preface, so the first bytes of a connection identify the protocol without
+// consuming anything. Everything else on the port is treated as one plain
+// HTTP/1.x request so that curl, load balancer health checks, and future
+// HTTP/1.1-only consumers such as Prometheus can share the RPC port. The
+// RPC/non-RPC split is decided by the request route, not the HTTP version:
+// this file only detects the framing.
+
+#include "dispatch.h"
+
+#include <chrono>
+#include <cstring>
+#include <string>
+#include <thread>
+
+#include "lupine_log.h"
+
+namespace {
+
+constexpr char kH2Preface[] = "PRI * HTTP/2.0\r\n\r\nSM\r\n\r\n";
+constexpr size_t kH2PrefaceLength = sizeof(kH2Preface) - 1;
+// Bounds how long a connection may sit unidentified. Lupine clients send the
+// preface immediately after connecting, so only a silent or dribbling peer
+// ever waits this long.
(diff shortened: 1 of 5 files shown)


dispatch.cpp
```diff
@@ -0,0 +1,186 @@
...
+  response += "content-length: " + std::to_string(body.size()) + "\r\n";
+  response += "connection: close\r\n\r\n";
+  if (include_body) {
+    response += body;
+  }
+  return response;
+}
+
+int serve_http1(lupine_socket_t fd, const rpc_http2_server_metadata *metadata,
+                std::chrono::steady_clock::time_point deadline) {
```
kevmo314: Is there a bug here? If I send an HTTP/1.1 header but do not send a body, I will hang the server for `deadline`, and I can DoS the server by repeatedly sending multiple clients.

Each request should be on a separate thread, or ideally something lighter weight than a thread, idk what nghttp2 uses. Similarly, if that's implemented we don't need a deadline. It's cheap to hang the thread and eventually the socket will die in the kernel, at which point `recv` will return.

mesutoezdil: i don't think it can hang the server, dispatch runs in the per-connection child (fork at server.cpp:469), so a stalled client only ties up its own child, same as a client that never sends the h2 preface today. the deadline actually makes these cheaper than rpc conns, those can idle forever. can drop the deadline if you prefer though.

kevmo314: Ahh ok, yeah let's drop the deadline, the rest of this PR looks good to me.

mesutoezdil: done

PR #752, opened 2026-09-11
Prefetch library loads from a per-application profile on a pipelined lane

kevmo314:
## Summary

Since #747 every `cuLibraryLoadData` is one round trip, but cudart still issues them strictly serially (a kernel misses, its fatbin is loaded, the handle is awaited, the next miss loads the next library), so the gpt benchmark pays 59 × RTT plus 59 cold uploads (111 MB) at 150 ms. This takes those loads off the critical path:

- **Per-application profile.** The first remote load reads `~/.cache/lupine/libraries/profiles/<key>` (`$XDG_CACHE_HOME` honoured; `LUPINE_LIBRARY_CACHE` overrides the directory, empty disables), listing the images the previous run loaded in order, with a content-addressed copy of each image under `images/<xxh64>`. `<key>` is `LUPINE_LIBRARY_PROFILE` or the XXH64 of `/proc/self/cmdline`, so `python train.py` and `python eval.py` keep separate load orders while sharing image files. Every successful remote load rewrites the profile (atomic temp+rename) and stores its image if absent; the image directory is trimmed oldest-first above `LUPINE_LIBRARY_CACHE_BYTES` (default 1 GiB).
- **Pipelined prefetch on its own lane.** A detached thread opens its own HTTP/2 lane, sends `cuCtxSetCurrent` for the app's context and then every profiled load back-to-back (the plain `cuLibraryLoadData` op, whose response carries the kernel snapshot since #747), and collects the responses in send order with the same request writer and response reader as the normal path. The snapshot is applied by `lupine_read_library` on the reading thread, so the prefetch thread sets its thread-local context hint to the app's context before collecting. The whole profile costs one RTT plus transfer instead of one RTT per library. A 4 MB in-flight window makes early entries claimable while later ones upload and bounds how far the burst queues ahead of the app's own requests in the shared socket (measured below).
- **Claim by content.** `cuLibraryLoadData` hashes the packed image; a pending entry with the same hash, kind, option bytes and identical image bytes (full `memcmp`, so XXH64 needs no collision resistance) is claimed once, waiting for its response if it has not arrived, and the owner/image bookkeeping runs on the caller thread. Anything else — no profile, differing options, mismatching bytes, prefetch failure — is a normal load.

Transport extension: `rpc_wait_for_pipelined_response(conn, write_id)` flushes (like `rpc_wait_for_response`, so a burst shares one LZ4 block per #751) and waits for an earlier request's response on the calling thread's lane. Responses on a lane arrive in send order (the server handles each stream sequentially), so a pipelining caller must collect them in that order before the lane waits for anything else; the prefetch thread owns its lane, so nothing else ever waits on it. `lupine_prepare_rpc` is deliberately called once at thread start rather than per request: the host flush it may issue is a synchronous RPC and must not interleave with the pipelined loads.

Invariants kept: every handle cudart receives is distinct (entries are consumed once, so `cuLibraryUnload` keeps its native meaning); a prefetched library belongs to the route it was loaded on and is claimed only for loads on that route; loads carrying JIT options are never recorded (their log/wall-time outputs belong to the call that passes them); library options are matched by wire bytes, so pointer-valued options that change between runs fall back. Only image files that hash to their own name are trusted, and a file that fails the check is removed so the run's normal load rewrites it.

Why not the alternatives: dedupe alone does nothing here (cudart never loads the same image twice per process). Pipelining on the app's lane would have to drain every outstanding prefetch response before any unrelated RPC on that lane could wait, blocking setup for the whole upload; a separate lane keeps the app's lane clean and lets setup overlap the transfer.

Diff: +448/−43 in `cuda_client.cpp`, `rpc.cpp`, `rpc.h`. No server or wire changes.

## Measured

node-006 (GPU-less client, `CUDA_VISIBLE_DEVICES=""`) → node-008 (RTX 4090), `gpt_bench_steps.py`. Server built from this branch (server code is identical to main). "main" = a client shim built from `origin/main` (`8d7faa4e`); "cold" = this branch with no profile (its per-step path is main's plus hashing and recording); "warm" = the following run.

150 ms RTT (netem, port 15311), NSTEPS=200 (main/cold rows at NSTEPS=20 for the per-step medians):

| | main | cold (no profile) | warm |
|---|---|---|---|
| setup | 19.76 s | 19.82 s | **15.38 s** |
| step 0 | 13.98 s | 14.02 s | **12.09 s** |
| step 1 | 2.72 s | 2.73 s | 3.03 s |
| steps 2+ median | 1967 ms | 1970 ms | 2271 ms |
| prefetch burst | — | — | 59/59 libraries in 6.3 s (111 MB) |

Note the steady state: current main runs ~2 s per training step at 150 ms (it was 6 ms on the #747 branch tip this PR was first measured on — setup 20.9 s, step 0 13.4 s, steps 2+ 5.98 ms). The pure-main shim against the same server reproduces it exactly, so it is a regression already on main (post #746/#751), independent of this change; it also inflates every synchronous RPC in step 0/1, which is why the step-0 gain here (−1.9 s) is smaller than on the #747 tip (13.41 → 10.94 s, −2.5 s, with setup 20.92 → 16.57 s).

LAN (~0.3 ms, port 15321), NSTEPS=200:

| | main | cold | warm |
|---|---|---|---|
| setup | 2.46 s | 2.47 s | 2.45 s |
| step 0 | 745 ms | 794 ms | **321 ms** |
| steps 2+ median | 9.5 ms | 11.2 ms | 11.5 ms |

(On the #747 tip the same comparison was 481–487 ms → 84–91 ms for step 0 with steps 2+ at 5.6–6.3 ms on both sides.)

Synchronous RPC counts (`LUPINE_RPC_STATS`, LAN): main/cold issue 30 `cuLibraryLoadData` round trips in setup and 29 in step 0. Warm: 0 synchronous loads on the app thread in setup and step 0 — all 59 are claimed from the prefetch lane (the op count of 59 is the prefetch thread's sends, 0 ms blocked). A profile written by a setup-only run (30 libraries) prefetches those 30 and loads the other 29 normally. All other per-phase counts are unchanged.

Window sweep at 150 ms on the #747 tip (in-flight bytes → setup / burst): 2 MB → 16.35 s / 7.1 s; 4 MB → 16.50 s / 6.1 s; 16 MB → 17.91 s / 5.2 s; 64 MB → 18.54 s / 5.0 s. Larger windows finish the burst sooner but every app-thread RPC during the burst queues behind more prefetch bytes in the shared TCP socket; 4 MB is the knee.

## Validation

- `ctest`: 12/12 passed (3 skipped, no local GPU) on the rebased branch; `.claude/fmt-check.sh` clean against main.
- Stale/corrupt profile (on the #747 tip): hand-edited a bogus hash line, appended a junk line, truncated one cached image and overwrote another with garbage. The run prefetched the remaining 56/56, loaded the 3 others normally, completed, removed the two bad image files and rewrote a full 59-entry profile.
- Losses: the bench is unseeded, so trajectories vary run to run; final losses are within the run-to-run spread, and the no-profile run — which never touches the prefetch path — sits in the same band as the warm runs.

## Caveats

- Only the route of the first remote load is prefetched; other routes of a multi-GPU app load normally.
- Profiled libraries the app no longer loads stay loaded on the server for the life of the connection (bounded by the 4096-entry profile cap and the cache size limit); the next run's profile drops them.
- On non-Linux clients there is no `/proc/self/cmdline`, so the profile is disabled unless `LUPINE_LIBRARY_PROFILE` names one.
- The remaining gap to "loads are free" is the upload itself (111 MB at ~18–21 MB/s over the 150 ms path, wire-bound) and the head-of-line delay it imposes on app-thread RPCs during the burst. A dedicated bulk connection or `TCP_NOTSENT_LOWAT` on the shared socket would remove the latter; compression of the images beyond the stream's lz4 would shrink the former.
- The unit tests do not exercise the prefetch (it needs a live server); behaviour was validated end to end on node-006/node-008 as above.

### DIFF
--- a/cuda_client.cpp
+++ b/cuda_client.cpp
@@ -2,14 +2,18 @@
 
 #include <algorithm>
 #include <atomic>
+#include <chrono>
+#include <condition_variable>
 #include <cstdint>
 #include <cstring>
 #include <cuda.h>
 #include <cuda_occupancy.h>
+#include <filesystem>
 #include <fstream>
 #include <functional>
 #include <iostream>
 #include <map>
+#include <memory>
 #include <mutex>
 #include <new>
 #include <sstream>
@@ -59,6 +63,7 @@
 #include "rpc.h"
 #include "third_party/libcuckoo/libcuckoo/cuckoohash_map.hh"
 #include "transport.h"
+#include "xxhash.h"
 
 #ifdef cuMemPrefetchAsync
 #undef cuMemPrefetchAsync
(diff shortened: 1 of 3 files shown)


kevmo314:
Closing as superseded by #881, which keeps the per-application profile and the single batch round trip but drops the client image cache, the pipelined lane and window, the transport extension, and the client-side hashing.

PR #788, opened 2026-09-15
Run CodeQL as an advanced-setup workflow that retries a failed results upload

kevmo314:
> **Required settings change (owner only):** switch CodeQL from default setup to advanced setup in **Settings > Code security > Code scanning > CodeQL analysis > ⋯ > Switch to advanced**, when this merges. While default setup is enabled, GitHub rejects uploads from an advanced-setup workflow, so this workflow's upload steps fail until the switch.

## Summary

Adds `.github/workflows/codeql.yml`, an advanced-setup equivalent of the repository's CodeQL default setup: same languages (`actions`, `c-cpp`, `python`), default query suite, build mode `none`, push and pull request to `main` plus a weekly schedule, and the same `Analyze (<language>)` job names. Analysis and upload are separate steps. If the upload to code scanning fails, the analysis SARIF is kept as a one-day artifact and a `Retry upload (<language>)` job uploads it again a minute later, without re-running the analysis. The retry has to be its own job: `upload-sarif` refuses a second upload for the same category in one job (`Aborting upload: only one run of the codeql/analyze or codeql/upload-sarif actions is allowed per job per tool/category`), which the first version of this PR hit. Runs of this workflow can also be re-run manually, unlike default-setup runs. When the first upload succeeds, the retry jobs find no artifact and do nothing.

## Root cause

Only 2 of the last 200 CodeQL runs (back to 2026-09-09) failed, both `Analyze (c-cpp)` on 2026-09-15, 30 minutes apart, and both failed at the same point:

| Run | Mode | Analysis | Upload |
|---|---|---|---|
| main 34952028892 (09:20) | improved incremental (`overlay-base`) | completed; `Exported results to SARIF`; 113/118 C++ and 100/119 C files scanned | `Uploading results`, then 39 s of silence, step failed with no upload error |
| PR #787 34954941465 (09:52) | incremental disabled (`none`) | completed; `Exported results to SARIF` | `Uploading results`, then 18 s of silence, step failed with no upload error |

Successful runs print `Successfully uploaded results` within about a second of `Uploading results`. No GitHub status incident covers either time, and re-pushing #787 with an identical tree passed.

- **Not disk space.** The main run's `##[error]` blaming disk space for improved incremental analysis is the generic post-step message the action prints whenever a job fails while that mode is on. Analysis had finished and the SARIF was written before the failure, and the #787 failure happened with the mode already off. The side effect is that the failure was recorded in the Actions cache (`codeql-overlay-status-cpp-2.27.0-runner-140GB`), so improved incremental analysis now stays off for CodeQL 2.27.0.
- **Not repository content or size.** The same scan of the same files succeeds before and after, and #787 passed unchanged on a second push.
- **Most likely a transient failure of the SARIF upload to code scanning.** Confidence: medium on where it failed (both logs stop inside the upload with analysis complete), low on why, because default setup logs no upload error and cannot be re-run with debug logging.

Default setup cannot retry an upload and its runs cannot be re-run, so each such blip leaves a red check until the next push. This workflow retries the upload and makes the job re-runnable.

Generated sources (`codegen/gen_*`) are deliberately still scanned: they are not implicated in the failures, and the generated server handlers are where marshalling memory bugs would live.

## Validation

- `actionlint` 1.7.7: clean.
- PR run 34958627845 (default setup still enabled): all three `Analyze` jobs completed analysis (c-cpp job 80 s end to end); each first upload was rejected with `Code Scanning could not process the submitted SARIF file: CodeQL analyses from advanced configurations cannot be processed when the default setup is enabled`, so each saved its SARIF artifact; each `Retry upload` job downloaded it, checked out the source, waited, and uploaded again (`Successfully uploaded results`), and code scanning rejected it for the same default-setup reason. The retry path therefore runs end to end; a successful upload can only be observed after the settings switch.
- After switching to advanced setup, the first push to `main` should show green `Analyze` jobs and `Retry upload` jobs that skip their steps.
- The Actions cache entry `codeql-overlay-status-cpp-2.27.0-runner-140GB` recorded the misattributed improved-incremental failure and keeps that mode off for CodeQL 2.27.0; deleting it (Actions > Caches) re-enables it. Not required for this fix.


kevmo314:
Closing: CodeQL C/C++ failed 2 of the last 200 runs, both transient upload failures, which is acceptable to ignore without moving off default setup.

Issue #832, opened 2026-09-18
cuFFTDx mixed_precision_fft_1d times out on 2.75M per-block synchronous HtoD copies


kevmo314:
`MathDx/cuFFTDx/08_mixed_precision/mixed_precision_fft_1d` is the one MathDx
library sample that does not finish through the shim. It is a round-trip count
problem, not a correctness one: the sample fills its input with one synchronous
`cudaMemcpy` **per CUDA block**, and it sizes the launch so that at least 1 GiB
of data is processed per FFT size.

## Evidence

Native on the same GPU (RTX 4090, CUDA 13.3, `inferable-node-006`), the sample
passes and reports its launch geometry:

```
FFT size: 128     FFTs per block: 1   CUDA blocks: 1048576
FFT size: 512     FFTs per block: 1   CUDA blocks: 262144
FFT size: 2048    FFTs per block: 1   CUDA blocks: 65536
Not enough resources for FFT of size: 4096
```

The upload loop is two copies per block
(`MathDx/cuFFTDx/08_mixed_precision/mixed_precision_fft_1d.cu`):

```c++
for (unsigned int i = 0; i < cuda_blocks; ++i) {
    cudaMemcpy(input_compute + i * single_fft_size, single_fft_host_input.data(),
               single_fft_size_compute_bytes, cudaMemcpyHostToDevice);
    cudaMemcpy(input_io + i * single_fft_size, single_fft_host_mixed_input.data(),
               single_fft_size_io_bytes, cudaMemcpyHostToDevice);
}
```

That is 2 x (1048576 + 262144 + 65536) = **2,752,512 blocking HtoD copies** of
1-4 KiB each, before the cuFFT reference run and the four cuFFTDx benchmark
launchers. Every one is a client-to-server round trip. At the 0.31 ms RTT
between the dev box and `inferable-node-006` that is ~850 s of pure latency, so
the unit hits `SAMPLE_TIMEOUT` (300 s in the run below, 120 s in the GPU
integration lane) long before it gets to the kernels.

Same host, same binaries, 127 MathDx units:

| | native | shim |
| --- | --- | --- |
| PASS | 127 | 126 |
| FAIL | 0 | 1 (`mixed_precision_fft_1d`, timeout) |

`mixed_precision_fft_2d` next to it passes: it runs one (512, 512) transform, so
it makes a handful of copies instead of millions.

## Notes

- Nothing is wrong on the wire. The sample re-uploads *the same* host buffer
[...]

PR #833, opened 2026-09-18
Enable the MathDx CUDALibrarySamples in the library-samples suite

### LINKED ISSUE #820: test: enable the MathDx CUDALibrarySamples in the library-samples suite

```
```

The header explains why the rest are out: they ship outside the CUDA toolkit. This issue tracks turning **MathDx** on, one library per issue so they can be enabled independently.

## To enable

Add `MathDx` to `LIBRARY_SAMPLES_LIBS`, make its headers and library available to the build the way `CUDNN_HOME` / `NCCL_HOME` / `CUSPARSELT_HOME` already are, and run the suite on a GPU node.

## Definition of done

## Shim status

**Possibly needs no shim.** MathDx (cuFFTDx, cuBLASDx, cuSolverDx) is device-side header-only; the host side should be plain cudart, which is already shimmed. Worth trying against the current shims before assuming work is needed — it may enable for free.

kevmo314:
Closes #820.

MathDx needed no shim. cuBLASDx, cuFFTDx, cuRANDDx, cuSolverDx and nvCOMPDx are
device-side: the package is headers plus `.fatbin`/`.a` device code linked into
the sample's own binary, and `nm -D --undefined-only` on the built samples finds
nothing at the host boundary but cudart, libcuda, NVRTC/nvJitLink and the
reference libraries (cuBLAS, cuBLASLt, cuFFT) the suite already covers:

```
cuBLASDx/01_gemm_introduction/introduction_example  libcublasLt.so.13 libcudart.so.13
cuBLASDx/15_nvrtc/nvrtc_gemm                        libcudart libnvrtc libcublas libcuda
cuFFTDx/06_convolution/convolution_performance      libcufft.so.12 libcudart.so.13
cuRANDDx/pcg_thread_api                             libcudart.so.13
nvCOMPDx/02_lz4_gpu/lz4_gpu_compression_decompression   (cudart_static; no NEEDED CUDA at all)
```

## Results

127 units on an RTX 4090 (`inferable-node-006`, CUDA 13.3, `LIBRARY_SAMPLES_ARCH=89`),
built from the same checkout and run natively on that GPU first, then through
the shim from a separate client host:

| | native | shim |
| --- | --- | --- |
| PASS | 127 | 126 |
| FAIL | 0 | 1 |

The one failure is `MathDx/cuFFTDx/08_mixed_precision/mixed_precision_fft_1d`,
now in `known_failures.txt` against #832. It fills >=1 GiB per FFT size with one
synchronous `cudaMemcpy` per CUDA block - 2,752,512 blocking HtoD round trips
before the first kernel - so it runs out the per-unit timeout on round-trip cost
alone. Nothing is wrong on the wire; it passes natively and `mixed_precision_fft_2d`
beside it passes through the shim.

## What the suite needed

- **The package.** MathDx ships outside the toolkit, and the release these
  samples require (cuBLASDx 0.7.1, cuFFTDx 1.7.3) is `nvidia-mathdx-26.06.1-cuda13`.
  The `nvidia-mathdx` PyPI wheel is not usable here: the newest is 25.6.0, which
  carries cuBLASDx 0.4.0 and no CMake package at all. The script unpacks the
  tarball next to the checkout the same way it clones the samples, so a run needs
  no new setup; `MATHDX_HOME` points at an existing copy instead.
- **The architecture.** Each library keeps its own `<LIB>_CUDA_ARCHITECTURES`
  cache variable, defaulted to `80-real` upstream, that `CMAKE_CUDA_ARCHITECTURES`
  does not reach. Left alone, every MathDx sample builds SM80-only cubins and
  dies with "no kernel image" on the sm_89 server. `build_sample` now derives the
  variable from the library directory name and sets it to the server's arch.
- **Project selection.** MathDx inverts the suite's leaf rule: one project per
  library builds all its examples, and the nested projects beneath it
  (`lto_helper`, `09_introduction_lto_example`, `10_cufft_device_api_example`)
  belong to `CUFFTDX_EXAMPLES_LTO`, off by default. The find filter keeps the
  library roots and drops the nested ones.
- **nvCOMPDx arguments.** Those samples take the file to compress on the command
  line; the four `unit_argv` lines mirror the arguments the nvCOMPDx CMakeLists
  passes to its own CTest entries, including the two random files its build
  generates.

## Port windows

`test/integration/CMakeLists.txt` gives each suite entry a ten-port window, and
the per-unit port was `SERVER_PORT_BASE + index` - fine when no sample built more
than ten executables, which every MathDx one does (cuSolverDx builds 42). The
index now wraps at ten, which is safe because units run one server at a time, and
keeps a sample inside its own window instead of walking into the next entry's.

## Cost

The five MathDx projects add about 15 minutes to a from-scratch sample build
(62 min CPU) and about 19 minutes to a suite run at this host's 0.31 ms RTT. The
slowest library stays well inside the 900 s CTest timeout: cuFFTDx 226 s,
cuBLASDx 362 s, cuSolverDx 229 s, nvCOMPDx 30 s, cuRANDDx 8 s.

## Full suite

All nine libraries, 197 samples, 358 units, 20 minutes:

```
PASS 343
FAIL 0
SKIP 15
TOTAL 358
```

The 15 skips are the 14 `known_failures.txt` entries that were already there
plus the MathDx one this PR adds. The eight libraries that were already enabled
are untouched.


kevmo314:
Two follow-up commits for the three GPU-lane failures.

## CUDA 12.4, both variants: MathDx 26.06 has no CUDA 12 build

`driver-only` failed the version check in `cusolverdx/detail/system_checks.hpp`;
`driver-runtime` failed earlier, on `cannot open source file "cstdint"`. Same
wall from two sides, and there is no CUDA 12 package to reach for: 26.06.x is
published for `cuda13` only (`.../cuda12/nvidia-mathdx-26.06.1-cuda12.tar.gz`
is a 404), and the newest release that does ship a CUDA 12 build, 25.12.1,
carries cuBLASDx 0.5.1 / cuFFTDx 1.6.1 / cuSolverDx 0.3.0 / cuRANDDx 0.2.2 --
below the 0.7.1 / 1.7.3 / 0.5.0 / 0.2.4 minimums each example project
`FATAL_ERROR`s on. `MathDx/nvCOMPDx/CMakeLists.txt` refuses CUDA < 13.0.0
outright.

It fails natively too. CUDA 12.4 container on `inferable-node-006`, same MathDx
package, no lupine in the path:

```
$ ./nvrtc_potrs
.../cusolverdx/detail/system_checks.hpp(22): catastrophic error: #error directive:
cuSolverDx requires NVRTC from CUDA Toolkit 13.0 or newer
exit=1
```

So the script now drops MathDx from `LIBRARY_SAMPLES_LIBS` below CUDA 13,
read from `CUDA_VERSION` in the toolkit's `cuda.h`. The 12.4 lane gets no
MathDx units instead of five build failures and one runtime failure.

The `cstdint` message did turn out to be a real shim defect, filed as #838:
`included_files()` marks a directive name known before it knows whether the
name resolves, so a bare name first sighted in a `#if`-disabled branch
suppresses a later resolvable occurrence from another directory. It is not what
gates this PR -- the sample cannot compile on CUDA 12 either way -- and on CUDA
13 nothing in the suite trips it. Twelve-line repro in the issue; native NVRTC
compiles it, the shim does not.

## CUDA 13.3: the two FP64 emulation samples run past the 120 s cap

`16_dgemm_emulation` and `20_dgemm_emulation_pipeline` both timed out at 122 s
in both variants. Not a hang and not the emulation path -- both finish, with
correct results. RTX 4090 on `inferable-node-006`, same binaries, 0.30 ms RTT:

| unit | native | shim |
| --- | ---: | ---: |
| `16_dgemm_emulation` | 56.5 s | 159.7 s |
| `20_dgemm_emulation_pipeline` | 44.0 s | 91.0 s |
| `21_sgemm_emulation_pipeline` | 9.6 s | 57.4 s |

The shim adds a near-constant ~47 s to both pipeline samples at this RTT, so
the FP64/FP32 split is the sample's own: each times an 8192^3 cuBLASLt DGEMM as
its reference against the emulated kernel, and the L4's FP64 rate is well under
the 4090's. The sgemm sibling passes on the lane at 24 s for the same reason it
is 9.6 s here.

That is #836's phenomenon, which already anticipated this pair ("MathDx also has
two units measured at 116-118 s on the test host"), so both are in
`known_failures.txt` against it.

PR #840, opened 2026-09-18
Look for an NVRTC include again when its name failed elsewhere

### LINKED ISSUE #838: NVRTC shim drops a header when an earlier unresolvable include claimed its name
The NVRTC shim resolves `#include`s on the client and ships the files it finds
to the server as named headers. `included_files()` in `nvrtc_client.cpp` marks a
directive name known **before** it knows whether the name resolves:

```c++
  std::set<std::string> known(state.include_names.begin(),
                              state.include_names.end());
...
      if (!known.insert(directive.first).second) {
        continue;
      }
      std::vector<std::string> candidates;
      ...
      for (const std::string &candidate : candidates) {
        ...
      }
```

So the first sighting of a bare name wins even when nothing was found for it.
A later `#include "name"` of the *same* bare name from a different directory --
which would resolve -- is skipped, the file never travels, and the server's
NVRTC fails to open it.

The walker deliberately does not evaluate conditional compilation ("a file that
is never used only costs its bytes"), which is what makes this reachable: the
poisoning sighting is usually inside a `#if`-disabled host-only branch that the
real preprocessor never takes.

## Repro

Three headers under `inc/`:

```
inc/first.hpp        #ifdef NEVER_DEFINED
                     #include "shared.h"
                     #endif
inc/sub/second.hpp   #include "shared.h"
inc/sub/shared.h     #define VAL 7
```

and a program compiled with `--include-path=<inc>`:

```c++
#include <first.hpp>
#include <sub/second.hpp>
__global__ void k(int *p) { *p = VAL; }
```

`first.hpp`'s `shared.h` resolves nowhere (it is not beside `first.hpp` and not
on the include path) but is recorded; `sub/second.hpp`'s resolves beside itself
and is dropped.

```
native NVRTC 13.3      result=OK
through the shim       sub/second.hpp(1): catastrophic error:
                       cannot open source file "shared.h"
```

Same binary, same headers, RTX 4090 server on `inferable-node-006`.

## Where it was found

Diagnosing #833's CUDA 12.4 lane. `MathDx/cuSolverDx/nvrtc_potrs` fails there
under `driver-runtime` with

```
detail/libcxx/include/type_traits(561): catastrophic error: cannot open source file "cstdint"
```

because `commondx/detail/stl/cstdint.hpp` sights the bare name first, in its
host-only `#else` branch:

```c++
#ifdef COMMONDX_DETAIL_USE_CUDA_STL
#    include <cuda/std/cstdint>
#else
#    include <cstdint>
#endif
```

and CUDA 12.4's `cuda/std/detail/libcxx/include/type_traits` then asks for it
as `#include "cstdint"`, which sits beside that file on disk. Native NVRTC 12.4
compiles `#include <cuda/std/type_traits>` fine; the shim cannot.

That sample fails on CUDA 12 for an unrelated reason as well (cuSolverDx 26.06
`#error`s below NVRTC 13.0, natively too), so #833 scopes MathDx out below CUDA
13 and this does not gate it. On CUDA 13 the CCCL layout does not route through
a quoted `"cstdint"`, so nothing in the suite trips it today -- it is latent,
not benign.

## Fix

Only record a name once a candidate was read, and keep the per-file walk from
retrying a name it already failed on in the same directory. A name that
resolves nowhere costs one directory probe per including file, which is
cheaper than losing the header.

kevmo314:
Fixes #838.

`included_files()` marked an include name known before it tried to resolve it,
so the first sighting of a bare name won even when nothing was found for it: a
later `#include "name"` of the same name from a directory that does have it was
skipped, the file never travelled, and the server's NVRTC could not open it.

The name now enters the set of names the server has only once a file is read.
A separate set remembers which lookups were already made, keyed by the
directory a quoted name was sighted from, so an unresolvable name costs one
failed lookup per directory that mentions it instead of hiding the resolvable
one. An angled name's candidates do not depend on the sighting, so it is still
looked for once.

`test_nvrtc_shim` grows a case for it: a header includes `"shared.h"` inside a
branch the preprocessor never takes and nothing resolves it, a second header in
another directory includes the same name with the file beside it, and the
kernel needs the macro that file defines. Without the change the compile fails
with `sub/second.h(1): catastrophic error: cannot open source file "shared.h"`.

The real chain is confirmed too: with CUDA 12.4's CCCL headers on the client and
an RTX 4090 server, `#include <cuda/std/type_traits>` fails through the shim on
`main` with `detail/libcxx/include/type_traits(561): catastrophic error: cannot
open source file "cstdint"` and compiles with the change. Under strace that walk
makes 57 failed candidate opens over the whole CCCL tree, so the retry stays
cheap.


kevmo314:
Closing: the NVRTC shim is being removed with the other runtime-library shims (epic #890, #889). Programs keep using NVIDIA's own NVRTC on the client, so the include-resolution path this fixes no longer exists. #838 will be closed with #889.

Issue #809, opened 2026-09-18
test: enable the cuTENSOR CUDALibrarySamples in the library-samples suite


kevmo314:
`test/run_cuda_library_samples.sh` builds and runs NVIDIA/CUDALibrarySamples through the lupine client shim against a remote server, but selects only a subset by default:

```
LIBRARY_SAMPLES_LIBS="cuBLAS cuBLASLt cuFFT cuRAND cuSOLVER cuSPARSE nvJPEG NPP"
```

The header explains why the rest are out: they ship outside the CUDA toolkit. This issue tracks turning **cuTENSOR** on, one library per issue so they can be enabled independently.

## To enable

Add `cuTENSOR` to `LIBRARY_SAMPLES_LIBS`, make its headers and library available to the build the way `CUDNN_HOME` / `NCCL_HOME` / `CUSPARSELT_HOME` already are, and run the suite on a GPU node.

Triage whatever fails the way the suite already does it (PRs #586/#588/#600): record each failure in `test/cuda-library-samples/known_failures.txt` and open one issue per distinct root cause. Compare against a native run on the same GPU before filing anything — a sample that fails natively is not a lupine bug.

## Definition of done

The library is in the default `LIBRARY_SAMPLES_LIBS`, its samples build and run through the shim, and every remaining failure is either in `known_failures.txt` with a linked root-cause issue or explained in the PR.

## Shim status

**No shim.** A `lupine_cutensor_client` has to be written first, following the pattern of the recent shim PRs (#792 nvJitLink is the cleanest template; `codegen/README.md` is required reading).

kevmo314:
## Assessment of #809-#824 — the remaining CUDALibrarySamples enablements

Posting here because there is no umbrella issue; this covers the whole set, not just cuTENSOR.

Measured on an RTX 4090 (`inferable-node-006`, CUDA 13.3, driver 590.48.01, `-arch=89`),
client on a separate host, **driver shim only** — `LD_PRELOAD=libcuda.so.1` with no
library shim of any kind in `LD_LIBRARY_PATH`, so every one of these runs the vendor's
own host library on the client and remotes only its driver traffic.

### The premise holds

**None of these needs a new shim.** Every host library tested so far runs unmodified on
the client over the driver shim. Three of the four failures found are not "the library
needs a shim" at all — they are a missing driver entry point, a driver-shim behaviour bug,
and a vendor-version mismatch in the samples. Only the `*Mp` set is structurally hard, and
[...]

PR #856, opened 2026-09-18
Enable the cuSOLVERSp2cuDSS CUDALibrarySamples in the library-samples suite

### LINKED ISSUE #813: test: enable the cuSOLVERSp2cuDSS CUDALibrarySamples in the library-samples suite

```
```

The header explains why the rest are out: they ship outside the CUDA toolkit. This issue tracks turning **cuSOLVERSp2cuDSS** on, one library per issue so they can be enabled independently.

## To enable

Add `cuSOLVERSp2cuDSS` to `LIBRARY_SAMPLES_LIBS`, make its headers and library available to the build the way `CUDNN_HOME` / `NCCL_HOME` / `CUSPARSELT_HOME` already are, and run the suite on a GPU node.

## Definition of done

## Shim status

**No shim**, and it is a migration guide from cuSOLVERSp to cuDSS, so it depends on the cuDSS shim. Enable after cuDSS.

kevmo314:
Closes #813.

cuSOLVERSp2cuDSS needed no shim. The samples run on the client against NVIDIA's
own `libcudss.so.0` and the toolkit's `libcusolver`; only their driver traffic
crosses the wire, which is what the `driver-only` loader variant already does for
`libcudart`. All seven units pass.

## The version boundary, which is the whole difficulty

These samples do **not** compile against cuDSS 0.8.0. They use `cudssAlgType_t`
and `CUDSS_ALG_1`, which 0.8.0 removed, and pass `cudaDataType_t` where 0.8.0
introduced a separate `cudssDataType_t` in a new `cudss_data_types.h`:

```
csreigs2cuDSS_dcomplex.cpp(183): error: identifier "cudssAlgType_t" is undefined
csreigs2cuDSS_dcomplex.cpp(183): error: identifier "CUDSS_ALG_1" is undefined
csreigs2cuDSS_dcomplex.cpp(229): error: argument of type "cudaDataType_t" is
                                incompatible with parameter of type "cudssDataType_t"
```

Built against each cuDSS release in turn:

| cuDSS | result |
| --- | --- |
| 0.6.0.5 | compile errors (an earlier API break), and no CUDA 13 build exists |
| 0.7.0.20 | builds clean |
| 0.7.1.4 | builds clean |
| 0.8.0.10 | the errors above |

Meanwhile all 22 `cuDSS/*/CMakeLists.txt` call `find_package(cudss 0.8.0
REQUIRED)`. **The two suites cannot share one cuDSS install**, so this one brings
its own under `CUDSS_TRANSITION_HOME`, pinned to 0.7.1.4, and leaves the cuDSS
samples' `CUDSS_HOME` (#812) alone. That is cheap because this project takes a
plain `-DCUDSS_ROOT` path rather than using `find_package`.

## What else the suite needed

- **The package.** cuDSS ships outside the toolkit; its redist archive carries
  `include/` and `lib/libcudss.so -> .so.0 -> .so.0.7.1`. The script unpacks it
  next to the checkout, keyed to the toolkit's own `CUDA_VERSION`; a lane whose
  major NVIDIA does not publish gets `SKIP:build-failed`.
- **`--solver cudss`.** With no arguments the samples run the *legacy*
  cuSOLVERSp/Rf path, which the cuSOLVER samples already cover; `--solver cudss`
  is the transition path these samples exist to demonstrate, so that is what the
  units run.

## Results

RTX 4090 on `inferable-node-006`, CUDA 13.3, cuDSS 0.7.1.4, `-arch=89`, client on
a separate host, run through `test/run_cuda_library_samples.sh` itself:

```
CUDALibrarySamples 3437729, 1 samples, 7 units
PASS 7
FAIL 0
SKIP 0
```

| unit | residual `|Ax-b|/|A|` |
| --- | --- |
| `cuSolverSp2cuDSS_float` | 4.6e-08 |
| `cuSolverSp2cuDSS_double` | 1.2e-16 |
| `cuSolverSp2cuDSS_scomplex` | 6.6e-08 |
| `cuSolverSp2cuDSS_dcomplex` | 1.4e-16 |
| `cuSolverRf2cuDSS_double` | 8.3e-17 |
| `csreigvsi2cuDSS_double` | eigenvalue 0.704567 |
| `csreigs2cuDSS_dcomplex` | 2 eigenvalues |

All seven also pass natively on the same GPU, in all three modes (default,
`--solver cudss`, and `--solver cudss -s`). Nothing to add to
`known_failures.txt`.

One thing to know if stdout comparison is ever added for these: the `scomplex`
and `dcomplex` residuals are nondeterministic **natively** - four consecutive
native runs gave 3.29e-06, 2.07e-06, 2.19e-06 and 2.38e-06 - so they would have
to be masked.

## Cost

One CMake project, 7 executables, about 90 s of build and 20 s of run.

Assessment of the whole #809-#824 set is in
[the comment on #809](https://github.com/lupinemachines/lupine/issues/809#issuecomment-5726593235).


kevmo314:
Rebased onto main, which now carries the cuDSS 0.8 samples (#854) too. Both installs ship `libcudss.so.0`, so the shared LD_LIBRARY_PATH now carries only the unit's own: `$CUDSS_TRANSITION_HOME/lib` for cuSOLVERSp2cuDSS units, `$CUDSS_HOME/lib` for everything else. With both on the path, whichever came first would have been loaded by both suites.

PR #868, opened 2026-09-18
Bind a lane whose thread's first CUDA call is a copy

### LINKED ISSUE #863: First CUDA call on a thread fails with invalid device context when it is an HtoD copy
A thread whose **first** CUDA call is a host-to-device copy fails with `invalid device
context`. Reproduced at 1 KiB and 1 MiB on plain `main` at `6048ace1`.

This is a **regression**: the same case passed before PR #835 ("device-binding epoch" /
lane binding, merged 2026-09-18). Found while building PR #839; that test works around it
with an explicit `cudaSetDevice` before the copy.

Worth confirming first, since the attribution is from bisection-by-observation rather than
a bisect: reproduce on `main`, then on #835's parent, and name the commit.

The likely shape, given what #835 changed: a lane with no prior CUDA call has no binding
recorded, and the copy path now consults that binding instead of establishing one. The
first call on a lane is exactly the case #835's own analysis identified as needing a
server round trip, so the copy path may be reading the binding without triggering the
bind.

This is on `main` today and affects any thread that starts by copying — a normal shape in
worker-thread code.

kevmo314:
Fixes #863.

## The attribution is wrong: #835 is innocent

The issue credits #835 from observation. Bisected instead, running the same
case on each candidate against `inferable-node-006` (RTX 4090) from a
GPU-less client:

| commit | driver-runtime |
| --- | --- |
| `e00f0723` (`main`) | FAIL `cudaErrorDeviceUninitialized` |
| `bb28bb3b` — #835's parent | FAIL, identically |
| `a83565e1` (#722, "Build and wire CUDART with paired integration variants") | FAIL, identically |
| `a83565e1^` | the case does not exist |

`a83565e1` is the commit that adds `cudart_client.cpp`, and its `cudaMemcpy`
has routed host-to-device copies straight into the driver shim's
`cuMemcpyHtoD` since that first line of it. There is no commit where the
driver-runtime variant exists and this passes, so this is not a regression
from this morning: it is a gap as old as the C++ runtime shim, three days
old. The earlier `python/cudart` shim was a separate object and is not this
code path.

It is also **driver-runtime only**. In `driver-only`, NVIDIA's own libcudart
does its lazy per-thread binding before it reaches `cuMemcpyHtoD`, so the
lane is already bound and the case passes on `main`.

## Mechanism

The server gives every HTTP/2 stream its own `std::thread` (`lupine_serve_lanes`),
and a client thread gets its own stream, so a client thread's first request
opens a lane no CUDA call has ever crossed — nothing is current on it.

The runtime shim's `cudaMemcpy` delegates to the driver shim, which resolves
its route with `lupine_route_for_deviceptr`. For a pointer this client already
tracks that returns the owning route straight out of the allocation table, and
unlike `lupine_route_for_default` it neither refreshes the runtime context nor
calls `lupine_set_current_context_on_route`. The synchronous copy carries no
stream either, so the driver has nothing to derive a context from and answers
`CUDA_ERROR_INVALID_CONTEXT` — `cudaErrorDeviceUninitialized`, "invalid device
context".

Two things the same probe settled:

* **The client cannot supply the context itself.** For a runtime-allocated
  pointer, `lupine_note_deviceptr_allocation` records
  `lupine_current_context_hint()`, which is null on a thread that has only
  made runtime calls: `cudaMalloc` is a forwarded runtime RPC, so the server's
  libcudart creates the context and the client never learns its address.
  Binding the allocation's own context is therefore not available without a
  round trip — the same fact #835 and #861 rest on.
* **`cudaMemcpyAsync` on an explicit stream already works**, because the
  driver services a stream-bearing call from the stream's context. The null
  stream does not.

## Fix

The runtime is what binds a thread: a runtime call on a thread with no current
context takes the current device's primary context. The driver shim must keep
failing a contextless driver-API caller, so the bind belongs in the runtime
shim, and it is `cudaSetDevice`, which has initialized the device's primary
context eagerly since CUDA 12. `bind_calling_lane` re-selects the device the
lane already reports, at the head of the copies that reach the driver shim.

It is skipped when the driver shim already has this thread on a context — the
steady state for anything touching the driver API, one thread-local read, and
also what keeps the re-selection from displacing a non-primary context the
application put on the lane. Failing that, a recorded `cudaSetDevice` under a
held binding epoch is the only other proof the lane is bound; a device query is
not, because `cudaGetDevice` is answered from that same record and the server
need not have been asked at all. Measured with `LUPINE_RPC_STATS`:
`test_bulk_copies` sends **zero** `cudaGetDevice` and `cudaSetDevice` either
way, and the new test sends exactly one pair per thread that copies — four
pairs for its four threads, however many copies each makes.

No new state, no new mechanism, and nothing in `routing.cpp`, `cuda_client.cpp`
or the memcpy files, so it does not collide with #861, #862 or #857. #862's
`lupine_demand_fetch_context` solves the adjacent case one layer down, on the
RPC dispatch thread's lane, where no runtime call is in flight to hang this on.

## Test

`test/test_thread_first_call_htod.cu`: a thread whose very first CUDA call is
a host-to-device copy, with no `cudaSetDevice` — at 1 KiB and 1 MiB, plus the
async form on an explicit stream, plus a thread that calls `cudaGetDevice`
before its copy.

* Mutation-tested: with `bind_calling_lane`'s body disabled and everything else
  in place, 3/3 runs fail all three synchronous cases with
  `cudaErrorDeviceUninitialized (201)`; 3/3 pass with it, and 2/2 driver-only.
* The `cudaGetDevice` case is why the skip needs `selected`: it failed against
  the first version of this fix, whose guard accepted any recorded binding.
* Custom tests, both loader variants: **27 tests × 2 = 54 runs, 0 failures**,
  including `test_runtime_device_query_cache` (#835's own), `test_thread_default_context`,
  `test_bulk_copies`, `test_pre_init_state`, `test_invalid_device_ordinal`,
  the HtoD/DtoH staging and ordering tests, the mapped-host tests, the graph
  and capture tests, and `test_unified_memory_coherence`.
* Unit tests 13/13 bar `hip_smemcpy_test`, which fails on this workstation's
  local AMD GPU and touches nothing here. clang-format clean on changed lines.
* Client on this workstation, server on `inferable-node-006` (RTX 4090), ports
  14846/14847.

## #839

Yes — #839's writer thread can drop its `cudaSetDevice` workaround once this
lands. Not touched here.


kevmo314:
Closing: this is a lane-binding fix inside the CUDA runtime shim (`cudart_client.cpp`), which is being removed (epic #890, #889). The driver shim's lane binding is unaffected by #863; the issue will be closed with #889.

Issue #869, opened 2026-09-18
A round-tripping runtime RPC can overtake a fire-and-forget launch on another lane


kevmo314:
`test_cross_lane_launch_order`'s `test_cross_type_order` only passes because the
`cudaGetLastError()` it calls after each launch round-trips. Delete that one line
and the test fails on `main`:

```
$ sed 's/check(cudaGetLastError(), "publish launch");//' test/test_cross_lane_launch_order.cu > /tmp/nobarrier.cu
$ nvcc --cudart=shared -arch=all /tmp/nobarrier.cu -o /tmp/nobarrier -lcuda -L$CUDA_HOME/lib64/stubs
$ LD_LIBRARY_PATH=<driver+runtime shims> LUPINE_SERVER=node:port /tmp/nobarrier
FAIL: 315 of 512 async memsets overtook kernel launches      # driver-runtime
PASS: async operations preserve order across client lanes    # driver-only
```

A kernel launch is fire-and-forget: `cudaLaunchKernel` takes an async sequence
number and the server gates native submission on it
(`rpc_async_sequence_begin`). `cudaMemsetAsync` in the runtime shim is an
ordinary round-tripping RPC with no sequence number, so nothing orders it
against a launch another lane issued first. The request that carries a 4 KiB
kernel-parameter payload is still on the wire when the small memset request
arrives on another stream, and the server submits the memset first.

The driver-only loader is unaffected because the driver shim marks
`cuMemsetD8/16/32Async` `@async`, so both operations go through the same
submission sequence.

Two shapes of fix:

* Mirror the driver shim: give the runtime shim's stream-ordered entry points
  (`cudaMemset*Async`, `cudaEventRecord*`, `cudaGraphLaunch`,
  `cudaLaunchHostFunc`, ...) the same `@async` submission the driver shim gives
  their driver twins. Also removes those round trips. Not a pure annotation
  change: `validate_async_annotation` rejects anything but a `CUresult` return,
  and `emit.py`'s client path only implements the optional (NCCL) flavour of
  `@async`, not the driver shim's unconditional fire-and-forget.
* Or make every request observe the sequence: stamp ordinary requests with the
  issuing lane's async watermark in the RPC frame header and have the server
  wait for `serving_async_sequence` to reach it before running the handler.
  This closes the whole class rather than the annotated subset, at the cost of
  a core wire-format change.

This is why the client-side `cudaGetLastError` change cannot merge as
[...]

PR #870, opened 2026-09-18
Answer the runtime's last-error queries on the client

kevmo314:
`cudaGetLastError` currently makes a round trip after each kernel launch and device-to-device copy checked by torch. Keep the runtime's last error in client thread-local storage so `cudaGetLastError` and `cudaPeekAtLastError` can answer locally. Runtime entry points record their returned failures; successful calls and `cudaErrorNotReady` leave the saved error unchanged. `cudaGetLastError` reads and clears it, while peek only reads it.

The generator records results directly for ordinary forwarders and stubs. Body templates with early returns retain an outer recorder so every return path is covered. Handwritten entry points use the same helper, without an additional wrapper when a single return suffices. Error-name and error-string queries remain unchanged.

This PR does not change the transport protocol or add cross-thread submission barriers. The added last-error CUDA test has been removed.

Validation:

- Rebuilt and passed all seven CPU suites after removing the watermark: transport, visible devices, checkpoint, HTTP/2, dispatch, events, and IPC. Whitespace checks pass.
- Regenerated sources with the pinned codegen container and verified all backends during the recording cleanup.
- Temporary last-error probes passed in both loader variants, covering sticky errors, peek/get behavior, thread isolation, query `NotReady`, and an asynchronous device fault. Generated and handwritten local validation failures also updated last-error state.
- CI is rerunning after removal of the connection-wide async watermark. Earlier green GPU results included that watermark and do not validate this revision.

Known ordering issue: #869 remains unresolved. The existing `test_cross_lane_launch_order` hands work between host threads using a shared CUDA stream; making `cudaGetLastError` local removes an incidental round-trip barrier and exposes that ordering failure in `driver-runtime`. Each individual transport lane continues to process requests sequentially.

Known error-reporting limitation: a launch rejected synchronously by the server's driver is still fire-and-forget, so its rejection does not reach the client's last-error state. This already occurs in `driver-only`; `driver-runtime` loses the recovery previously provided by querying the server's runtime state. Device faults still surface through subsequent synchronizing calls.

The original 150 ms WAN benchmark reduced the 500-step training loop from 380.4 s to 79.1 s and round trips per step from 5.07 to 1.05. These measurements preceded the wrapper cleanup.


kevmo314:
Closing: the CUDA runtime shim is being removed (epic #890, sources deleted in #889), so a client-side last-error cache for it is moot. The measurement in this PR is preserved in the epic as part of the evidence.

PR #879, opened 2026-09-21
Enable the nvTIFF CUDALibrarySamples in the library-samples suite

### LINKED ISSUE #816: test: enable the nvTIFF CUDALibrarySamples in the library-samples suite

```
```

The header explains why the rest are out: they ship outside the CUDA toolkit. This issue tracks turning **nvTIFF** on, one library per issue so they can be enabled independently.

## To enable

Add `nvTIFF` to `LIBRARY_SAMPLES_LIBS`, make its headers and library available to the build the way `CUDNN_HOME` / `NCCL_HOME` / `CUSPARSELT_HOME` already are, and run the suite on a GPU node.

## Definition of done

## Shim status

**No shim.** 8 sample directories.

kevmo314:
Closes #816.

nvTIFF needed no shim. The samples run on the client against NVIDIA's own
`libnvtiff.so.0`; only its driver traffic crosses the wire, which is what the
`driver-only` loader variant already does for `libcudart`. All 6 units pass,
and every file they write is byte-identical to a native run on the same GPU.

## What the suite needed

- **The package.** nvTIFF ships outside the toolkit. Its redist archive carries
  `include/nvtiff.h` and `lib/libnvtiff.so -> .so.0 -> .so.0.8.0.82`, which is
  the layout the samples' `find_library(... HINTS ${NVTIFF_PATH}/lib)` looks
  for, so nothing has to be staged. The script unpacks it next to the checkout
  the same way it does the other redist archives; `NVTIFF_HOME` points at an
  existing copy instead. The archive is CUDA-major-specific, so the URL is
  built from the toolkit's own `CUDA_VERSION` and a lane whose major NVIDIA
  does not publish gets `SKIP:build-failed`.
- **0.8.0.82.** The samples are written against the 0.8 API (upstream README:
  "nvTIFF 0.8 or later"); it is the newest release and is published for CUDA 12
  and 13.
- **Arguments.** Every sample takes its input TIFF on the command line and exits
  on its usage message without one. The samples ship a single striped image,
  `nvTIFF/images/bali_notiles.tif`, which the decode samples take directly.
  `nvTIFF-Image-Info-Multi-Image` wants a multi-page file and
  `nvTIFF-GeoTIFF-Decode-Encode` wants georeferencing; `nvTIFF-Encode-Options`
  writes both (`striped_lzw_multipage.tif`, and `custom_tags_from_buffer.tif`
  with `ModelPixelScale`/`ModelTiePoint`), and it sorts before them, so they
  consume its output the way the nvJPEG2000 encoder consumes the decoder's.
  Outputs go to `$TMPDIR/lupine-nvtiff-assets` like the nvJPEG/nvJPEG2000 ones.
- nvTIFF `dlopen`s nvJPEG (and nvJPEG2000/nvCOMP) on demand for those codecs;
  the JPEG variant in `nvTIFF-Encode-Options` finds `libnvjpeg` through the
  `CUDA_LIB_DIR` already on the unit's `LD_LIBRARY_PATH`.

## Results

RTX 4090 on `inferable-node-008` (driver 590.48.01, device 0 as the server
enumerates it), client CUDA 13.3, nvTIFF 0.8.0.82, `-arch=89`, client on a
separate host, run through `test/run_cuda_library_samples.sh` itself with
`LIBRARY_SAMPLES_LIBS=nvTIFF`:

```
CUDALibrarySamples 3437729, 6 samples, 6 units
PASS 6
FAIL 0
SKIP 0
```

The samples do not check pixels themselves, so the same binaries were also
run natively on the same GPU: the 10 decoded PPM previews and the 7 TIFFs
written by the encode samples have identical md5s through lupine and natively.

No skips. One limitation of the fixture: `custom_tags_from_buffer.tif` carries
the model tags but no GeoKeyDirectory, so the GeoTIFF sample reports
`0 geo key(s)` and does not exercise the geo-key setters. Those are host-side
metadata calls; its GPU decode and encode path runs either way. A file with
geo keys needs GDAL or a hand-written TIFF, neither of which the suite has.

## Cost

6 small CMake projects, under 20 seconds of build (the whole run, clone to
summary, was about 30 seconds) and 9 seconds of unit time. The archive is
2.4 MB. No unit comes close to `SAMPLE_TIMEOUT`.


kevmo314:
CI runs each library sample as its own parallel ctest, so the unit-order dependency this PR relied on didn't hold there: nvTIFF-GeoTIFF-Decode-Encode started before nvTIFF-Encode-Options had written its input and failed with NVTIFF_STATUS_INVALID_PARAMETER (CUDA 12.4 and 13.3 L4 lanes). The added commit gives the consumers a ctest DEPENDS on their producer, and a per-variant NVTIFF_ASSETS_DIR/NVJPEG2K_ASSETS_DIR so driver-only and driver-runtime don't share one /tmp dir. nvJPEG2000-Encoder had the same latent dependency on nvJPEG2000-Decoder from #849, so it's covered too.

PR #880, opened 2026-09-21
Enable the NPP+ CUDALibrarySamples in the library-samples suite

### LINKED ISSUE #817: test: enable the NPP+ CUDALibrarySamples in the library-samples suite

```
```

The header explains why the rest are out: they ship outside the CUDA toolkit. This issue tracks turning **NPP+** on, one library per issue so they can be enabled independently.

## To enable

Add `NPP+` to `LIBRARY_SAMPLES_LIBS`, make its headers and library available to the build the way `CUDNN_HOME` / `NCCL_HOME` / `CUSPARSELT_HOME` already are, and run the suite on a GPU node.

## Definition of done

## Shim status

**No shim.** NPP Plus is a separate C++ API library (`libnppplus`), not covered by the eleven `lupine_npp*_client` shims.

kevmo314:
Closes #817.

NPP+ needed no shim, and it does not go through the `lupine_npp*` shims either.
`libnpp_plus_{if,c,isu}.so.0` link no toolkit NPP and no shared cudart: they
carry a static CUDA 12 runtime that dlopens `libcuda.so.1`, which resolves to
lupine's driver shim. The samples' own `cudaMalloc`/`cudaMemcpy` calls go
through lupine's `libcudart.so.13`. `LD_DEBUG=files` on a unit shows the
libraries it loads: lupine's `libcuda.so.1` and `libcudart.so.13`, the two
`libnpp_plus` libraries and libc/libstdc++. No `libnpp*` is loaded. 4 of the 5
units pass; the fifth segfaults natively too.

## What the suite needed

- **The package.** NPP+ ships outside the toolkit. Its redist archive carries
  `include/` (NPP+'s own `npp.h` plus `nppPlus/`) and `lib/` with the usual
  symlink chain. The script unpacks it next to the checkout the same way it
  does the others; `NPPPLUS_HOME` points at an existing copy instead.
- **CUDA 11 and 12 only.** 0.10.0.0 is the newest release, and NVIDIA
  publishes it only for `cuda11` and `cuda12`, as both the redist archive and
  the `libnppplus0-*-cuda-1x` debs. The URL is still built from the toolkit's
  `CUDA_VERSION`, but it is capped at 12. The archive brings its own runtime and
  needs only a driver, so the CUDA 12 build is the right one for a CUDA 13
  toolkit too. That is how it was run here.
- **The CMake lists assume the deb layout.** They search for the libraries
  under `${NPP_PLUS_PATH}/12` and for `npp.h` under `/usr/include/libnpp_plus/12`.
  The archive has neither path. The script passes the five cache variables
  directly (`NPP_HEADER_PATH`, `NPP_PLUS_HEADER_PATH` and the three
  `NPP_PLUS_*_LIBRARY`), so `npp.h` resolves to NPP+'s copy and not the
  toolkit's.
- **A macro mismatch upstream.** The sources call `nppPlusV::...`, and
  `nppPlus.h` declares that namespace only under `NPP_PLUS_ENABLE`. The lists
  define `NPP_PLUS_ON`, so all five samples fail to compile against the
  published headers. The redist and deb headers are byte-identical, so this is
  not a packaging difference. The script adds `-DNPP_PLUS_ENABLE`. The symbols
  are `extern "C"`, so the namespace does not change linkage.
- **Working directories.** The inputs are opened relative to the working
  directory, with the same layout as the NPP samples: `findContour` runs from
  the sample dir and the rest run from `images/`. `distanceTransform` is the
  exception. It builds its paths as `"../images" + name`, with no separator, and
  none of the five files it names ship in its own directory; they are in
  `watershedSegmentation/images/`. It runs from a staging directory whose
  parent holds symlinks under the names it builds. Without that, every load
  fails silently and the sample exits 0 having done nothing: a hollow PASS.

## Results

RTX 4090 on `inferable-node-006` (driver 590.48.01), CUDA 13.3 toolkit, NPP+
0.10.0.0 (cuda12 archive), `-arch=89`, client on a separate host, run through
`test/run_cuda_library_samples.sh` itself from a clean download and build:

```
CUDALibrarySamples 3437729, 5 samples, 5 units
PASS 4
FAIL 0
SKIP 1
```

What the passing units produced:

- `distanceTransform` writes 16 transform outputs. All 16 are byte-identical to
  a native run of the same binary on the same GPU.
- `watershedSegmentation` writes 12 outputs. All 12 are byte-identical to the
  native run and to the reference files checked into the samples repo.
- `batchedLabelMarkersAndCompression` and `floodFill` print compressed label
  counts and fill bounds that differ between runs. They also differ between
  native runs: four native runs back to back gave four different sets, for
  example PCB `1466/1465/1466/1466` and CT skull `435/437/436/436`. The UF
  labelling is nondeterministic on this GPU, and the shim's numbers fall inside
  the native spread.

The one skip:

- **`findContour`** -> `known_failures.txt`, in a native-failure block with no
  lupine issue. It segfaults natively on every run (3 of 3) with the same
  binary on the same GPU. The crash is in the sample's own `main`, not in the
  library. It walks each contour to `nMaxContourPixelCount - 1` in unsigned
  arithmetic, so a count of 0 becomes about 4G iterations of writes through
  garbage coordinates.

### CUDA 11 archive (CI's 11.8 lane)

The 11.8 lane fetches the **cuda11** archive, and
`batchedLabelMarkersAndCompression` failed there in both loader variants with
`LabelMarkersUFBatch_8Way_8u32u failed.` To reproduce it, that sample was built
in `nvidia/cuda:11.8.0-devel-ubuntu22.04` against the cuda11 archive and run on
the same RTX 4090:

| run | result |
|---|---|
| native, in the 11.8 container | PASS, 3/3 |
| through a lupine client + server built with CUDA 13.3 | PASS, 3/3 |
| through a lupine client + server built with CUDA 11.8, driver-only | FAIL, 2/2 |
| through a lupine client + server built with CUDA 11.8, driver-runtime | FAIL, 2/2 |

The binary and the libraries are the same in every row, so this is lupine, and
it depends on the toolkit lupine is built with, not on the NPP+ archive. The
runtime NPP+ bundles looks up `cuFuncGetName`, `cuKernelGetName` and
`cuStreamGetDevice` through `cuGetProcAddress`. An 11.8-built client compiles
all three out under `CUDA_VERSION` guards, and the runtime then launches no
kernel at all: `LUPINE_TRACE` shows zero `cuLaunchKernel` calls, against 60+ from
the 13.3 build. The sample is now in `known_failures.txt` against **#882**.

The same cause affects the other NPP+ units on the 11.8 lane even though CI
reports them as passing. `watershedSegmentation`, `floodFill` and
`distanceTransform` check only NPP status codes, and NPP+ returns success
without having launched anything. On 11.8, `watershedSegmentation` writes
all-0/1 segment maps instead of the reference output. Their PASS on that lane is
hollow until #882 is fixed; the details are on the issue. The CUDA 12/13 results
above are real passes, confirmed by comparing the outputs byte for byte.

One lupine defect was found: #882, which affects CUDA 11.8 builds only.

## Cost

The archive is 365 MB compressed and about 950 MB unpacked. The samples link
three of its libraries, 180 MB. There are 5 CMake projects. On this host the
download, unpack and build took about 35 s and the run about 15 s. No unit
comes near `SAMPLE_TIMEOUT`.

**Update:** the #882 defect found here is fixed by #883 (merged). This PR no longer lists `batchedLabelMarkersAndCompression` in `known_failures.txt`; with #883 the CUDA 11.8 lane runs all NPP+ units for real (batchedLabelMarkersAndCompression, distanceTransform, floodFill, watershedSegmentation pass in both loader variants; findContour stays skipped as a native failure).


kevmo314:
Rebased onto #883 (the #882 fix) with the #882 known_failures.txt entry dropped, so the CUDA 11.8 GPU integration lane runs NPP+/batchedLabelMarkersAndCompression through an 11.8-built lupine that forwards cuFuncGetName/cuKernelGetName/cuStreamGetDevice. Once #883 merges this rebases back onto main as the single NPP+ commit.

